PR #2, opened 2018-11-08
Adding Circle CI.

Squirrel18:
## Description:

This PR adds the CircleCI config file.
No run jobs were added, since there is not test to run yet, it is just the initial configuration.

## Reviewers:

 - [x] @felipemontoya 
 - [ ] @Ivanca 
 - [ ] @diegomillan


felipemontoya:
I think this is good, but we need to add the tests on every branch if we merge this first. If we merge the branches first, then we need to add the tests here

PR #9, opened 2018-11-28
Edge case on dicts initialized on top of strings

felipemontoya:
Fixes the case where a given key exists as a string on the base_settings and a dict on the tenant

### DIFF
--- a/eox_tenant/signals.py
+++ b/eox_tenant/signals.py
@@ -53,7 +53,10 @@ def _update_settings(domain):
     LOG.debug("PID: %s CONFIGURING THE SETTINGS OBJECT | %s", os.getpid(), tenant_key)
     for key, value in six.iteritems(config):
         if isinstance(value, dict):
-            merged = getattr(base_settings, key, {}).copy()
+            try:
+                merged = getattr(base_settings, key, {}).copy()
+            except AttributeError:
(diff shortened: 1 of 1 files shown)


eox_tenant/signals.py
```diff
@@ -53,7 +53,10 @@ def _update_settings(domain):
     LOG.debug("PID: %s CONFIGURING THE SETTINGS OBJECT | %s", os.getpid(), tenant_key)
     for key, value in six.iteritems(config):
         if isinstance(value, dict):
-            merged = getattr(base_settings, key, {}).copy()
+            try:
+                merged = getattr(base_settings, key, {}).copy()
+            except AttributeError:
```
Squirrel18: It is not necessary to add an attribute error, because if the attribute does not exist, getattr always returns the default value. @felipemontoya

felipemontoya: It is true that getattr returns the default, but the error we are catching here comes from the fact that not always `base_settings.key` will have a .copy attribute

felipemontoya: I have found cases where the production data will override a string with a dict

Squirrel18: Ok, i didn't realize that.

PR #10, opened 2018-11-28
Adding a new field to the admin page

felipemontoya:
This PR makes it easy to see on the admin which sites have activated signals

### DIFF
--- a/eox_tenant/admin.py
+++ b/eox_tenant/admin.py
@@ -16,11 +16,13 @@ class MicrositeAdmin(admin.ModelAdmin):
         'sitename',
         'template_dir',
         'course_org_filter',
+        'ednx_signal',
     ]
     readonly_fields = (
         'sitename',
         'template_dir',
         'course_org_filter',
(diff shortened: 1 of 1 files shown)


eox_tenant/admin.py
```diff
@@ -54,5 +56,15 @@ def course_org_filter(self, microsite):
...
 
+    def ednx_signal(self, microsite):
+        """
+        Read only method to see if the site has activated the usage of signals
+        """
+        # pylint: disable=broad-except
+        try:
+            return microsite.values.get('EDNX_USE_SIGNAL', "EMPTY")
+        except Exception as error:
+            return str(error)
```
Squirrel18: We can catch a NameError exception, since get returns a None by default and values returns an empty list, so the only known error that could be raised is an NameError, if microsite is not defined for some reason.

felipemontoya: We have a broad in all methods of this page because we want to make sure that nothing prevents the page from rendering

PR #18, opened 2018-12-11
BC-21 | Custom microsite methods

felipemontoya:
Moving custom microsite implementation over from core platform

### DIFF
--- a/eox_tenant/backends/database.py
+++ b/eox_tenant/backends/database.py
@@ -1,6 +1,9 @@
 """
 Microsite backend that reads the configuration from the database
 """
+import os
+
+from django.conf import settings
 from django.utils import six
 from eox_tenant.backends.base import BaseMicrositeBackend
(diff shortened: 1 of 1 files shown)


eox_tenant/backends/database.py
```diff
@@ -151,3 +160,29 @@ def _set_microsite_config_from_obj(self, subdomain, domain, microsite_object):
...
+                microsites_root
+            )
+
+    def set_key_to_cache(self, key, value):
+        """
+        Stores a key value pair in a cache scoped to the thread
+        """
+        if hasattr(self.current_request_configuration, 'cache'):
+            self.current_request_configuration.cache[key] = value
+        if not hasattr(self.current_request_configuration, 'cache'):
```
Squirrel18: Think this could be an elif instead of if not? @felipemontoya

felipemontoya: Nice catch. This logic actually looks weird now. I'm going to review it in more detail.

PR #20, opened 2018-12-14
HA/EDNX/FFI-3 - Add if statement to fix lms tests.

Squirrel18:
## Description:

This PR adds an if statement to filter_enrollments function to verify if the function was called during the test process.

This is because some certificates test are failing due to the filter_enrollments function makes more database calls than normal behaviour.

## Reviewers:

 - [ ] @diegomillan 
 - [ ] @felipemontoya

### DIFF
--- a/eox_tenant/tenant_aware_functions/enrollments.py
+++ b/eox_tenant/tenant_aware_functions/enrollments.py
@@ -1,6 +1,9 @@
 """
 Microsite aware enrollments filter.
 """
+
+from django.conf import settings
+
 from eox_tenant.edxapp_wrapper.get_microsite_configuration import get_microsite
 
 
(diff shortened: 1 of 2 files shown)


eox_tenant/tenant_aware_functions/enrollments.py
```diff
@@ -10,6 +13,10 @@ def filter_enrollments(enrollments):
     do not belong to the current microsite
     """
 
+    if getattr(settings, "EOX_TENANT_AWARE_FUNCTION_FILTER_ENROLLMENTS", False):
```
felipemontoya: We can make this be one the boolean checks. The second would be the `not get_microsite().is_request_in_microsite()` and we can use the function already on L21

felipemontoya: Also, the setting could be way clearer: `EOX_TENANT_SKIP_FILTER_FOR_TESTS`

felipemontoya:
@Squirrel18 please squash before merging

felipemontoya:
Also, can you set the `EOX_TENANT_SKIP_FILTER_FOR_TESTS` variable from this repo?
we do have a test.py, but it might not be enough. Can you check locally?

Squirrel18:
Ok @felipemontoya let me check it out.

Squirrel18:
It works locally @felipemontoya i am waiting for running test, and then i going to merge it.

PR #22, opened 2018-12-18
Adding command to modify tenants over a cli.

felipemontoya:
This PR adds a new edit_tenant_values django command.

The command supports adding and removing nested keys and using basic interpolation with the tenant values. Also supports pattern filtering to select the sites it will operate on.

### DIFF
--- a/eox_tenant/management/commands/edit_tenant_values.py
+++ b/eox_tenant/management/commands/edit_tenant_values.py
@@ -1,0 +1,160 @@
+"""
+This module contains the command class to modify the values
+from a tenant so that we can add, remove of edit tenants in bulk.
+"""
+from __future__ import unicode_literals
+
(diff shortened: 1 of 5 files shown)


eox_tenant/management/commands/edit_tenant_values.py
```diff
@@ -0,0 +1,160 @@
+"""
+This module contains the command class to modify the values
+from a tenant so that we can add, remove of edit tenants in bulk.
+"""
+from __future__ import print_function, unicode_literals
```
Squirrel18: I think print is not being used. @felipemontoya

felipemontoya: You are right. I was using it everywhere but migrated to LOGGER at the end

PR #28, opened 2019-02-22
Reloading apps as configured in the tenant

felipemontoya:
This PR allows a particular tenant to reload the AppConfig from a INSTALLED_APP after the settings are reset.

This  fixes the issue we had on the ThirdPartyAuth sad paths.
The method called here: https://github.com/eduNEXT/edunext-platform/blob/master/common/djangoapps/third_party_auth/apps.py#L15 overrides some settings that where reset on the signal call.

We allow a tenant to decide if it wants any app being reloaded again using `EDNX_TENANT_INSTALLED_APPS`

### DIFF
--- a/eox_tenant/signals.py
+++ b/eox_tenant/signals.py
@@ -25,6 +25,8 @@ import os
 from datetime import datetime
 
 import six
+
+from django.apps.config import AppConfig
 from django.conf import settings as base_settings
 
 from eox_tenant.backends.database import EdunextCompatibleDatabaseMicrositeBackend
@@ -71,8 +73,6 @@ def _repopulate_apps(apps):
     We delegate the decision to the specific tenant on the EDNX_TENANT_INSTALLED_APPS key.
     If present, the key is passed in the apps argument
     """
-    from django.apps.config import AppConfig
-
     LOG.debug("PID: %s REPOPULATING APPS | %s", os.getpid(), apps)
     for entry in apps:
         app_config = AppConfig.create(entry)
--- a/eox_tenant/signals.py
+++ b/eox_tenant/signals.py
@@ -63,6 +63,22 @@ def _update_settings(domain):
         setattr(base_settings, key, value)
 
 
+def _repopulate_apps(apps):
+    """
+    After the initial loading of the settings, some djangoapps can override the AppConfig.ready() method
+    to alter the settings in a particular way.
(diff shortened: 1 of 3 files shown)


felipemontoya:
I'll be writing some tests now.

jfavellar90:
My only question is: why some apps reload is required?

felipemontoya:
In particular third party auth applies some settings on top of the bare settings. During the signal we reset to the base settings and we lose the thirdpartyauth extra.
I could not find any other app doing this, but I rather make the feature reusable

jfavellar90:
Just deleted my comment because the answer was specified on the PR description

eox_tenant/signals.py
```diff
@@ -63,6 +63,22 @@ def _update_settings(domain):
...
 
+def _repopulate_apps(apps):
+    """
+    After the initial loading of the settings, some djangoapps can override the AppConfig.ready() method
+    to alter the settings in a particular way.
+    In this case we need to run the app config again.
+    We delegate the decision to the specific tenant on the EDNX_TENANT_INSTALLED_APPS key.
+    If present, the key is passed in the apps argument
+    """
+    from django.apps.config import AppConfig
```
jfavellar90: importing this on the top of the file could cause any issue?

eox_tenant/signals.py
```diff
@@ -176,6 +192,13 @@ def start_tenant(sender, environ, **kwargs):  # pylint: disable=unused-argument
     # Do the update
     _update_settings(domain)
 
+    # Some django apps need to be reinitialized
+    try:
+        if base_settings.EDNX_TENANT_INSTALLED_APPS:
```
jfavellar90: At this point we're playing with tenant settings because _update_settings was already called right?

eox_tenant/signals.py
felipemontoya: I don't think it will cause an issue. I was only trying to make it load only when the tenant requires it.
But that logic is flawed since this way it will have to load again and again instead of just one time when the server starts.
I'll fix this now

eox_tenant/signals.py
felipemontoya: correct

eox_tenant/signals.py
felipemontoya: Done

PR #29, opened 2019-06-14
Adding tenant config business logic

diegomillan:
This PR allows handling configs for multiple sites while moving our tenant support to another model. Tenant is only in charge of reading & loading the whole set of configurations.  

@felipemontoya 
@jfavellar90 
@Squirrel18


eox_tenant/models.py
```diff
@@ -65,6 +66,59 @@ def get_microsite_for_domain(cls, domain):
...
+class TenantConfig(models.Model):
+    """
+    Model to persist edxapp configurations.
+    """
+    key = models.CharField(max_length=63, db_index=True)
+    values = JSONField(null=False, blank=True, load_kwargs={'object_pairs_hook': collections.OrderedDict})
+    lms_configs = JSONField(null=False, blank=True, load_kwargs={'object_pairs_hook': collections.OrderedDict})
+    studio_configs = JSONField(null=False, blank=True, load_kwargs={'object_pairs_hook': collections.OrderedDict})
+    theming_configs = JSONField(null=False, blank=True, load_kwargs={'object_pairs_hook': collections.OrderedDict})
+    sites = models.ManyToManyField(Site)
```
felipemontoya: Why is this a ManyToMany? one site (domain) will not have many configurations. Having a many to many requires one extra lookup in a tenant-sites table instead of placing the foreign_key of the site in this table.

eox_tenant/models.py
```diff
@@ -65,6 +66,59 @@ def get_microsite_for_domain(cls, domain):
...
+
+    class Meta:
+        """
+        Model meta class.
+        """
+        app_label = "eox_tenant"
+
+    def __unicode__(self):
+        return self.key
```
felipemontoya: It will help you by a lot to also return the domain here

eox_tenant/models.py
```diff
@@ -65,6 +66,59 @@ def get_microsite_for_domain(cls, domain):
...
+        """
+
+        # remove any port number from the hostname
+        domain = domain.split(':')[0]
+        config = TenantConfig.objects.filter(sites__domain=domain).last()  # pylint: disable=no-member
+
+        if config:
+            return config.values, config.key
+
+        microsite = Microsite.get_microsite_for_domain(domain)
```
felipemontoya: it is weird to have this backwards compatibility here. I would rather put it on the `get_tenant_config_by_domain` method of the backends.database module

eox_tenant/models.py
```diff
@@ -65,6 +66,59 @@ def get_microsite_for_domain(cls, domain):
...
+        Get edxapp configuration using a domain. There is a compat layer to support microsite until
+        deprecation.
+        """
+
+        # remove any port number from the hostname
+        domain = domain.split(':')[0]
+        config = TenantConfig.objects.filter(sites__domain=domain).last()  # pylint: disable=no-member
+
+        if config:
+            return config.values, config.key
```
felipemontoya: Should this return the config.values? or better the config.lms_configs ?

eox_tenant/models.py
```diff
@@ -65,6 +66,59 @@ def get_microsite_for_domain(cls, domain):
         return microsites[0] if microsites else None
 
 
+class TenantConfig(models.Model):
+    """
+    Model to persist edxapp configurations.
+    """
+    key = models.CharField(max_length=63, db_index=True)
+    values = JSONField(null=False, blank=True, load_kwargs={'object_pairs_hook': collections.OrderedDict})
```
felipemontoya: why do we keep the values field. This should be replaced by lms_configs and completely remove values since it causes a lot of confusion with the old microsites.

eox_tenant/models.py
```diff
@@ -65,6 +66,59 @@ def get_microsite_for_domain(cls, domain):
...
+    lms_configs = JSONField(null=False, blank=True, load_kwargs={'object_pairs_hook': collections.OrderedDict})
+    studio_configs = JSONField(null=False, blank=True, load_kwargs={'object_pairs_hook': collections.OrderedDict})
+    theming_configs = JSONField(null=False, blank=True, load_kwargs={'object_pairs_hook': collections.OrderedDict})
+    sites = models.ManyToManyField(Site)
+
```
felipemontoya: What about the metada field we had discussed?

eox_tenant/models.py
```diff
@@ -65,6 +66,99 @@ def get_microsite_for_domain(cls, domain):
...
+    Custom managaer for Tenant Config model.
+    """
+    def get_configurations(self, domain):
+        """
+        Execute optimized query to get site configurations.
+        """
+        from django.db import connection
+
+        with connection.cursor() as cursor:
+            cursor.execute("""
```
jfavellar90: This with django queryset tools was an expensive operation?

eox_tenant/backends/persistent.py
```diff
@@ -0,0 +1,34 @@
+"""
+Site configuration backend that reads the configuration from the database
+"""
+
+
+class CompatibleDatabaseSiteConfigBackend(object):
```
jfavellar90: Is this a completely new backend unlinked from microsite backends?

eox_tenant/admin.py
```diff
@@ -67,6 +67,79 @@ def ednx_signal(self, microsite):
...
+
+    def sitename(self, tenant_config):
+        """
+        Read only method to calculate sitename attribute from config model.
+        """
+        # pylint: disable=broad-except
+        try:
+            return tenant_config.lms_configs.get('SITE_NAME', "NOT CONFIGURED")
+        except Exception as error:
+            return str(error)
```
jfavellar90: If you are returning str(error), this means that this error is going to be displayed in the django admin console?

eox_tenant/models.py
diegomillan: Yes, basically this is an optimized query based on https://www.wix.engineering/post/scaling-to-100m-mysql-is-a-better-nosql

eox_tenant/backends/persistent.py
diegomillan: Yes, but for now this backend reads/loads configurations. The second phase of the development will handle more operations.

eox_tenant/admin.py
diegomillan: it will be displayed on the django admin model interface. Specifically, when listing TenantConfig instances.

jfavellar90: and is this the behavior we want?

diegomillan: Well, this is something that should not happen. But if so, we don't want the whole admin list failing

eox_tenant/admin.py
```diff
@@ -67,6 +67,79 @@ def ednx_signal(self, microsite):
             return str(error)
 
 
+class TenantConfigAdmin(admin.ModelAdmin):
+    """
+    Tenant config model admin
```
Squirrel18: Please, add a period at the end of the line. @diegomillan

eox_tenant/admin.py
```diff
@@ -67,6 +67,79 @@ def ednx_signal(self, microsite):
...
+    )
+    search_fields = ('key', 'route__domain', 'lms_configs', )
+
+    def sitename(self, tenant_config):
+        """
+        Read only method to calculate sitename attribute from config model.
+        """
+        # pylint: disable=broad-except
+        try:
+            return tenant_config.lms_configs.get('SITE_NAME', "NOT CONFIGURED")
```
Squirrel18: Please, don't mix single quotes and double quotes. @diegomillan

eox_tenant/admin.py
```diff
@@ -67,6 +67,79 @@ def ednx_signal(self, microsite):
...
+        except Exception as error:
+            return str(error)
+
+    def template_dir(self, tenant_config):
+        """
+        Read only method to calculate template dir attribute from config model.
+        """
+        # pylint: disable=broad-except
+        try:
+            return tenant_config.lms_configs.get('template_dir', "NOT CONFIGURED")
```
Squirrel18: Please, don't mix single quotes and double quotes. @diegomillan

eox_tenant/admin.py
```diff
@@ -67,6 +67,79 @@ def ednx_signal(self, microsite):
...
+        except Exception as error:
+            return str(error)
+
+    def course_org_filter(self, tenant_config):
+        """
+        Read only method to calculate course org filter attribute from config model.
+        """
+        # pylint: disable=broad-except
+        try:
+            return tenant_config.lms_configs.get('course_org_filter', "NOT CONFIGURED")
```
Squirrel18: Please, don't mix single quotes and double quotes. @diegomillan

eox_tenant/admin.py
```diff
@@ -67,6 +67,79 @@ def ednx_signal(self, microsite):
...
+        except Exception as error:
+            return str(error)
+
+    def ednx_signal(self, tenant_config):
+        """
+        Read only method to see if the site has activated the usage of signals.
+        """
+        # pylint: disable=broad-except
+        try:
+            return tenant_config.lms_configs.get('EDNX_USE_SIGNAL', "EMPTY")
```
Squirrel18: Please, don't mix single quotes and double quotes. @diegomillan

eox_tenant/admin.py
```diff
@@ -67,6 +67,79 @@ def ednx_signal(self, microsite):
...
+
+    def domains(self, tenant_config):
+        """
+        Read only method to calculate the domain.
+        """
+        # pylint: disable=broad-except
+        try:
+            domains = [route.domain for route in tenant_config.route_set.all()]
+            separator = '\n'
+            return separator.join(domains)
```
Squirrel18: The separator may be avoided, since it's only used in the .join function. @diegomillan

eox_tenant/backends/persistent.py
```diff
@@ -0,0 +1,34 @@
+"""
+Site configuration backend that reads the configuration from the database
```
Squirrel18: Please, add a period at the end of the line. @diegomillan

eox_tenant/backends/persistent.py
```diff
@@ -0,0 +1,34 @@
...
+"""
+
+
+class CompatibleDatabaseSiteConfigBackend(object):
+    """
+    Backend that reads the configurations definitions from the database
+    using the custom models from edunext
```
Squirrel18: Please, add a period at the end of the line. @diegomillan

eox_tenant/backends/persistent.py
```diff
@@ -0,0 +1,34 @@
...
+    def _get_microsite_config_by_domain(self, domain):
+        """
+        Return the configuration and key available for a given domain.
+        """
+        from eox_tenant.models import Microsite
+        microsite = Microsite.get_microsite_for_domain(domain)
+
+        if microsite:
+            return microsite.values, microsite.key
+        else:
```
Squirrel18: This else is unnecessary after return. @diegomillan

eox_tenant/models.py
```diff
@@ -65,6 +66,99 @@ def get_microsite_for_domain(cls, domain):
         return microsites[0] if microsites else None
 
 
+class TenantConfigManager(models.Manager):
+    """
+    Custom managaer for Tenant Config model.
+    """
```
Squirrel18: Please, add a line break here. @diegomillan

eox_tenant/models.py
```diff
@@ -65,6 +66,99 @@ def get_microsite_for_domain(cls, domain):
...
+    def get_configurations(self, domain):
+        """
+        Execute optimized query to get site configurations.
+        """
+        from django.db import connection
```
Squirrel18: Is this to avoid a circular import? Since in the line 8 is importing from django.db too. @diegomillan

eox_tenant/models.py
```diff
@@ -65,6 +66,99 @@ def get_microsite_for_domain(cls, domain):
...
+
+        with connection.cursor() as cursor:
+            cursor.execute("""
+                SELECT eox_c.id, eox_c.external_key, eox_c.lms_configs, eox_c.studio_configs, eox_c.theming_configs, eox_c.meta
```
Squirrel18: Could you break each item in a new line? Just to make the query more readable. @diegomillan

eox_tenant/models.py
```diff
@@ -65,6 +66,99 @@ def get_microsite_for_domain(cls, domain):
...
+                FROM eox_tenant_tenantconfig eox_c
+                WHERE eox_c.id = (SELECT eox_tenant_route.config_id from eox_tenant_route
+                WHERE eox_tenant_route.domain=%s)""", [domain])
+            result_list = []
```
Squirrel18: Please, add a line break here and the 87 line. @diegomillan

eox_tenant/models.py
```diff
@@ -65,6 +66,99 @@ def get_microsite_for_domain(cls, domain):
...
+        deprecation.
+        """
+
+        # remove any port number from the hostname
+        domain = domain.split(':')[0]
+        config = TenantConfig.objects.filter(route__domain=domain).last()  # pylint: disable=no-member
+
+        if config:
+            return config.lms_configs, config.external_key
+        else:
```
Squirrel18: This else is unnecessary after return. @diegomillan

eox_tenant/models.py
```diff
@@ -65,6 +66,99 @@ def get_microsite_for_domain(cls, domain):
...
+            return {}, None
+
+    objects = TenantConfigManager()
+
+
+class Route(models.Model):
+    """
+    Model to persist site routes
```
Squirrel18: Please, add a period at the end of the line. @diegomillan

eox_tenant/test/test_backends.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/python
+"""
+TODO: add me
```
Squirrel18: Please, add it. @diegomillan

eox_tenant/test/test_backends.py
```diff
@@ -0,0 +1,28 @@
...
+"""
+from django.test import TestCase
+
+from eox_tenant.backends.database import EdunextCompatibleDatabaseMicrositeBackend
+from eox_tenant.backends.persistent import CompatibleDatabaseSiteConfigBackend
+
+
+class BackendsTest(TestCase):
+    """
+    Test backends methods
```
Squirrel18: Please, add a period at the end of the line. @diegomillan

eox_tenant/admin.py
jfavellar90: Here are we expecting to have a different exception than AttributeError? considering the case when lms_configs is not defined. On the other hand, can we have a default value in lms_configs like an empty dict?

eox_tenant/models.py
```diff
@@ -65,6 +67,115 @@ def get_microsite_for_domain(cls, domain):
...
+                SELECT
+                    eox_config.id,
+                    eox_config.external_key,
+                    eox_config.lms_configs,
+                    eox_config.studio_configs,
+                    eox_config.theming_configs,
+                    eox_config.meta
+                FROM eox_tenant_tenantconfig eox_config
+                WHERE eox_config.id = (SELECT eox_tenant_route.config_id from eox_tenant_route
+                WHERE eox_tenant_route.domain=%s)""", [domain])
```
felipemontoya: is this safe for SQL injection?

diegomillan: Yes, it is safe. Check references here:
https://docs.djangoproject.com/en/dev/topics/db/sql/#executing-custom-sql
https://docs.djangoproject.com/en/dev/ref/models/expressions/#django.db.models.expressions.RawSQL

eox_tenant/models.py
```diff
@@ -65,6 +67,115 @@ def get_microsite_for_domain(cls, domain):
...
+
+            # Using fetchone since the query will return one configuration per domain at the most.
+            row = cursor.fetchone()
+            if row:
+                configurations = {
+                    "id": row[0],
+                    "external_key": row[1],
+                    "lms_configs": json.loads(row[2]),
```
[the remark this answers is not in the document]
diegomillan: Feedback applied @felipemontoya

diegomillan:
> I believe we can squash this into a few commits as needed and we are ready.

Done  

ps: I will merge so we can test this on stage @felipemontoya

PR #30, opened 2019-07-03
adding route model admin

diegomillan:
This PR adds a django admin interface to handle Route model.

@jfavellar90 
@felipemontoya 
@Squirrel18


diegomillan:
Im merging since it's just a small change

PR #32, opened 2019-07-13
Ir/ednx/ha 3

morenol:
In this PR:

- Are removed the PathRedirectionMiddleware and RedirectionsMiddleware, they will live now in eox-core https://github.com/eduNEXT/eox-core/pull/71.
- It is created the AvailableScreenMiddleware which previously was a part of the RedirectionsMiddleware. 
- Created the theming helper backend

**Things to consider:**

Previously in the RedirectionsMiddleware (Now in the AvailableScreenMiddleware ) we were using  `microsite.is_request_in_microsite()` to detect if a request was part of a tenant.

Now we use theming_helper.is_request_in_themed_site() which in edx-platform is defined as:
    `return configuration_helpers.is_site_configuration_enabled() or microsite.is_request_in_microsite()`

Which means that this would work for both microsites and site configurations with a few aspects that need to be taken in consideration for the site configurations mode:

- In the django-sites, there is always a default site which is defined in the settings as **SITE_ID** (in edx-platform its default is 1).
- When a request does not match with any site, it will use the site defined with the SITE_ID configuration.
- If that default site has a site configuration enabled, then the available screen middleware will not work as expected.

For that reason, in this PR is included a way to change the default site in order to be capable of use the AvailableScreenMiddleware if the site with id 1 already has a site configuration


eox_tenant/settings/common.py
```diff
@@ -4,6 +4,7 @@
 
 # Quick-start development settings - unsuitable for production
 # See https://docs.djangoproject.com/en/1.11/howto/deployment/checklist/
+from path import Path as path
```
jfavellar90: Is this the reason why we had to add a new dependency in requirements @morenol ?

morenol: Yes

PR #33, opened 2019-07-18
Improving a bit the change_domain command

jfavellar90:
This PR allows to control in a better way how the stage domains are build. Specifically in the case where the provided suffix domains starts with a dot, for example ".i.stage.edunext.co". In this case we don't need a hyphen (-) as separator of the transformed microsite domain and the suffix.

@morenol 
@cocococosti 
@felipemontoya

### DIFF
--- a/eox_tenant/management/commands/change_domain.py
+++ b/eox_tenant/management/commands/change_domain.py
@@ -68,8 +68,11 @@ class Command(BaseCommand):
         my-microsite-domain-{suffix_stage_domain}
         """
 
+        pre_formatted = "{}-{}"
+        if self.suffix_stage_domain.startswith("."):
+            pre_formatted = "{}{}"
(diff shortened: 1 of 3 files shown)


morenol:
I think that we need to do the changes to that command to be compatible with site configurations

jfavellar90:
@morenol what you mean exactly? compatible with openedx Site Configurations? or with out new Tenancy model (introduced in version 0.6.0)?

morenol:
I mean, we are only changing the domains in the microsites and not in the django sites: https://github.com/eduNEXT/eox-tenant/blob/27919f95ee48d72e23e6f37685f3e56da953498e/eox_tenant/management/commands/change_domain.py#L41

jfavellar90:
@morenol I agree with that

PR #34, opened 2019-08-01
Use configuration helpers instead of microsites

morenol:
Change to use configuration helpers instead of microsites

### DIFF
--- a/eox_tenant/templatetags/ednx.py
+++ b/eox_tenant/templatetags/ednx.py
@@ -24,7 +24,7 @@ def favicon_path(default=getattr(settings, 'FAVICON_PATH', 'images/favicon.ico')
     Django template tag that outputs the configured favicon:
     {% favicon_path %}
     """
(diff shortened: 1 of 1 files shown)


jfavellar90:
@morenol this one should be the version 1.0.1 version I think

eox_tenant/templatetags/ednx.py
```diff
@@ -34,14 +34,14 @@ def microsite_css_overrides_file():
     Django template tag that outputs the css import for a:
     {% microsite_css_overrides_file %}
     """
-    file_path = microsite.get_value('css_overrides_file', None)
+    file_path = configuration_helpers.get_value('css_overrides_file', None)
```
morenol: I think this line is not necessary. @jfavellar90  What do you think?

jfavellar90: @morenol agree with it, let's procees to merge this one

PR #35, opened 2019-08-02
Jhony/fix filter enrollments

jfavellar90:
This PR mitigates the high cost that implies calling the get_all_orgs function from our microsite backend. Basically there are two changes:

- The method filter_enrollments does not call get_all_orgs method from the microsite backend unless it's strictly necessary.
- The result of the method get_all_orgs from the microsite backend is cached during 5 minutes

@diegomillan 
@felipemontoya 
@anmrdz 
@Squirrel18

### DIFF
--- a/eox_tenant/backends/database.py
+++ b/eox_tenant/backends/database.py
@@ -3,10 +3,15 @@ Microsite backend that reads the configuration from the database
 """
 import os
 
+from django.core.cache import cache
 from django.conf import settings
 from django.utils import six
+
 from eox_tenant.backends.base import BaseMicrositeBackend
 from eox_tenant.edxapp_wrapper.get_common_util import strip_port_from_host
+
+MICROSITES_ALL_ORGS_CACHE_KEY = 'microsites.all_orgs_list'
+MICROSITES_ALL_ORGS_CACHE_KEY_TIMEOUT = 300  # In seconds
 
 
 class EdunextCompatibleDatabaseMicrositeBackend(BaseMicrositeBackend):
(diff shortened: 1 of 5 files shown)


eox_tenant/tenant_aware_functions/enrollments.py
```diff
@@ -22,18 +22,23 @@ def filter_enrollments(enrollments):
...
+        orgs_to_include = []
+        orgs_to_exclude = get_microsite().get_all_orgs()
 
     for enrollment in enrollments:
 
         org = enrollment.course_id.org
 
         # Filter out anything that is not attributed to the inclusion rule.
-        if orgs_to_include and org not in orgs_to_include:
+        if org not in orgs_to_include:
```
felipemontoya: By looking at it, the logic looks the same to me, but, do we have a test that lets us be sure we are not creating a bug here?

eox_tenant/backends/database.py
```diff
@@ -3,11 +3,16 @@
...
 
+from django.core.cache import cache
 from django.conf import settings
 from django.utils import six
+
 from eox_tenant.backends.base import BaseMicrositeBackend
 from eox_tenant.edxapp_wrapper.get_common_util import strip_port_from_host
 
+MICROSITES_ALL_ORGS_CACHE_KEY = 'microsites.all_orgs_list'
+MICROSITES_ALL_ORGS_CACHE_KEY_TIMEOUT = 300  # In seconds
```
felipemontoya: should we read this value from settings so that we can tune it later on?

jfavellar90:
@felipemontoya added tests and setting to tune the all_orgs timeout

PR #41, opened 2019-08-13
[WIP] Ir/ednx/ir 20

morenol:
This PR adds an AuthenticationBackend that prevents users that signed up on a tenant site to login in other sites.

It also adds the capability to restrict the login in a tenant site to the users with an email that matches with the REGISTRATION_EMAIL_PATTERNS_ALLOWED pattern

This changes are based on:

https://github.com/eduNEXT/edunext-platform/commits/a52131f0a631398649eeb8241e89743feb18136f/common/djangoapps/student/views

### DIFF
--- a/eox_tenant/auth.py
+++ b/eox_tenant/auth.py
@@ -1,0 +1,82 @@
+
+"""
+This file implements the authentication backend for the openedx platform.
+"""
+import logging
+import re
+
+from django.conf import settings
+from django.contrib.auth import get_user_model
+
+from django.utils.translation import ugettext as _
+
+from eox_tenant.edxapp_wrapper.auth import get_edx_auth_backend, get_edx_auth_failed
+
+AuthFailedError = get_edx_auth_failed()
+AUDIT_LOG = logging.getLogger("audit")
+EdxAuthBackend = get_edx_auth_backend()
+UserModel = get_user_model()
+
+
+class TenantAwareAuthBackend(EdxAuthBackend):
+    """
+    Authentication Backend class which will check if the user has a signupsource in the requested site.
+    """
+    def authenticate(self, request, **kwargs):
+        """
+        Override default method to store the request which is being used by the
+        user_can_authenticate_on_tenant method.
+        """
+        self.request = request  # pylint: disable=attribute-defined-outside-init
+        return super(TenantAwareAuthBackend, self).authenticate(
+            request=request, **kwargs
+        )
+
+    def user_can_authenticate(self, user):
+        """
+        Override user_can_authenticate method with the logic to be tenant aware.
+        """
+        if not settings.FEATURES.get("EDNX_TENANT_AWARE_AUTH", False):
+            return super(TenantAwareAuthBackend, self).user_can_authenticate(user)
+
+        return self.user_can_authenticate_on_tenant(user)
+
+    def user_can_authenticate_on_tenant(self, user):
+        """
+        Prevent users that signed up on a different tenant site to login in this site.
+        """
+        current_domain = self.request.META.get("HTTP_HOST")
+
+        authorized_sources = getattr(settings, 'EDNX_ACCOUNT_REGISTRATION_SOURCES', [current_domain])
+        sources = user.usersignupsource_set.all()
+
+        is_authorized = False
+        for source in sources:
+            if any(re.match(pattern + "$", source.site) for pattern in authorized_sources):
+                is_authorized = True
+
+        email = getattr(user, 'email', None)
+        if settings.REGISTRATION_EMAIL_PATTERNS_ALLOWED is not None:
+            # Taken from forms.AccountCreationForm
+            allowed_patterns = settings.REGISTRATION_EMAIL_PATTERNS_ALLOWED
+            if not any(re.match(pattern + "$", email) for pattern in allowed_patterns):
+                # This email is not on the whitelist of allowed emails.
+                is_authorized = False
+
+        if not is_authorized:
+            loggable_id = user.id if user else "<unknown>"
+            if settings.FEATURES.get('EDNX_ENABLE_STRICT_LOGIN', False):
+                AUDIT_LOG.warning(
+                    u"User `%s` tried to login in site `%s`, but was denied permission based on the signup sources.",
+                    loggable_id,
+                    current_domain,
+                )
+                raise AuthFailedError(_('User not authorized to perform this action'))
+            else:
+                AUDIT_LOG.warning(
+                    u"User `%s` tried to login in site `%s`, the permission "
+                    "should have beed denied based on the signup sources.",
+                    loggable_id,
+                    current_domain,
+                )
+        return True
--- a/eox_tenant/edxapp_wrapper/auth.py
+++ b/eox_tenant/edxapp_wrapper/auth.py
@@ -1,0 +1,24 @@
+"""
+Auth definitions.
+"""
+
+from importlib import import_module
+from django.conf import settings
+
+
+def get_edx_auth_backend():
+    """ Get Edx Authentication Backend model. """
+
+    backend_function = settings.EOX_TENANT_EDX_AUTH_BACKEND
+    backend = import_module(backend_function)
+
+    return backend.get_edx_auth_backend()
+
+
+def get_edx_auth_failed():
+    """ Get the AuthFailed Class. """
+
+    backend_function = settings.EOX_TENANT_EDX_AUTH_BACKEND
+    backend = import_module(backend_function)
+
+    return backend.get_edx_auth_failed()
--- a/eox_tenant/settings/aws.py
+++ b/eox_tenant/settings/aws.py
@@ -4,6 +4,8 @@ Settings for eox_tenant project meant to be called on the edx-platform/*/envs/aw
 
 from .common import *  # pylint: disable=wildcard-import
 
+EDX_AUTH_BACKEND = 'openedx.core.djangoapps.oauth_dispatch.dot_overrides.backends.EdxRateLimitedAllowAllUsersModelBackend'  # pylint: disable=line-too-long
+
 
 def plugin_settings(settings):  # pylint: disable=function-redefined
     """
@@ -31,6 +33,10 @@ def plugin_settings(settings):  # pylint: disable=function-redefined
         'MICROSITES_ALL_ORGS_CACHE_KEY_TIMEOUT',
         settings.MICROSITES_ALL_ORGS_CACHE_KEY_TIMEOUT
     )
+    settings.EOX_TENANT_EDX_AUTH_BACKEND = getattr(settings, 'ENV_TOKENS', {}).get(
+        'EOX_TENANT_EDX_AUTH_BACKEND',
+        settings.EOX_TENANT_EDX_AUTH_BACKEND
+    )
 
     # Override the default site
     settings.SITE_ID = getattr(settings, 'ENV_TOKENS', {}).get(
@@ -43,3 +49,8 @@ def plugin_settings(settings):  # pylint: disable=function-redefined
             'eox_tenant.middleware.AvailableScreenMiddleware',
             'eox_tenant.middleware.MicrositeCrossBrandingFilterMiddleware',
         ]
+
+        if EDX_AUTH_BACKEND in settings.AUTHENTICATION_BACKENDS:
+            settings.AUTHENTICATION_BACKENDS.remove(EDX_AUTH_BACKEND)
+
+        settings.AUTHENTICATION_BACKENDS.append('eox_tenant.auth.TenantAwareAuthBackend')
--- a/eox_tenant/settings/aws.py
+++ b/eox_tenant/settings/aws.py
@@ -5,6 +5,7 @@ Settings for eox_tenant project meant to be called on the edx-platform/*/envs/aw
 from .common import *  # pylint: disable=wildcard-import
 
 EDX_AUTH_BACKEND = 'openedx.core.djangoapps.oauth_dispatch.dot_overrides.backends.EdxRateLimitedAllowAllUsersModelBackend'  # pylint: disable=line-too-long
+EOX_TENANT_AUTH_BACKEND = 'eox_tenant.auth.TenantAwareAuthBackend'
 
 
 def plugin_settings(settings):  # pylint: disable=function-redefined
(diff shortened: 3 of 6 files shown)


eox_tenant/settings/aws.py
```diff
@@ -43,3 +47,5 @@ def plugin_settings(settings):  # pylint: disable=function-redefined
             'eox_tenant.middleware.AvailableScreenMiddleware',
             'eox_tenant.middleware.MicrositeCrossBrandingFilterMiddleware',
         ]
+
+        settings.AUTHENTICATION_BACKENDS = ['eox_tenant.auth.TenantAwareAuthBackend'] + list(settings.AUTHENTICATION_BACKENDS)  # pylint: disable=line-too-long
```
morenol: @jfavellar90 I am not sure how I should do this. 
I only need to replace the 'openedx.core.djangoapps.oauth_dispatch.dot_overrides.backends.EdxRateLimitedAllowAllUsersModelBackend' with 'eox_tenant.auth.TenantAwareAuthBackend'

eox_tenant/auth.py
```diff
@@ -0,0 +1,84 @@
...
+        authorized_sources = getattr(settings, 'EDNX_ACCOUNT_REGISTRATION_SOURCES', [current_domain])
+        sources = user.usersignupsource_set.all()
+
+        is_authorized = False
+        for source in sources:
+            if any(re.match(pattern + "$", source.site) for pattern in authorized_sources):
+                is_authorized = True
+
+        email = getattr(user, 'email', None)
+        if settings.REGISTRATION_EMAIL_PATTERNS_ALLOWED is not None:
```
morenol: I added this here, but maybe this should be on eox-core

eox_tenant/auth.py
```diff
@@ -0,0 +1,95 @@
...
+AuthFailedError = get_edx_auth_failed()
+AUDIT_LOG = logging.getLogger("audit")
+UserModel = get_user_model()
+
+
+class TenantAwareAuthBackendBase(UserModelBackend):
+    """
+    Authentication Backend class which will check if the user has a signupsource in the requested site.
+    """
+    def authenticate(self, request, username=None, password=None, **kwargs):
```
jfavellar90: @morenol I think we can use the `super` method here to avoid replicating the code from django. What we can do is overwriting the user_can_authenticate method and there

morenol: @jfavellar90 the user_can_authenticate method does not receive the "request" which is being used to get the HTTP_HOST

jfavellar90: @morenol I think we can use the super method here to avoid replicating the code from django. What we can do is overwriting the user_can_authenticate method and there we can call the user_can_authenticate_on_tenant method to be a bit more compliant with the django backend structure

jfavellar90: @morenol in fact the only thing we need is to ovwerwrite the method `user_can_authenticate`

morenol: That was my first approach, but the user_can_authenticate method does not receive the request which is being used to get the HTTP_HOST

https://github.com/django/django/blob/bc1f79d0a01d085500aa82ff29800403291a91a4/django/contrib/auth/backends.py#L109

eox_tenant/auth.py
```diff
@@ -0,0 +1,95 @@
...
+import logging
+import re
+
+from django.conf import settings
+from django.contrib.auth import get_user_model
+from django.contrib.auth.backends import AllowAllUsersModelBackend as UserModelBackend
+
+from django.utils.translation import ugettext as _
+
+from ratelimitbackend.backends import RateLimitMixin
```
jfavellar90: would we need to add this dependency to eox-tenant dependencies? The inheritance didn't work the way we expected by using the `EdxRateLimitedAllowAllUsersModelBackend` class?

eox_tenant/auth.py
jfavellar90: @morenol one option could be overwriting `authenticate` method and in the first line make something like `self.request = request`, from that point it's available on any method, but I don't know what do you think about it

morenol: That is something that we can do, but I don't know if that is better than the current change

eox_tenant/auth.py
morenol: Yes, with the current changes, we need to add that dependency.

eox_tenant/edxapp_wrapper/auth.py
```diff
@@ -0,0 +1,15 @@
+"""
+Auth definitions.
+"""
+
+from importlib import import_module
+from django.conf import settings
+
+
+def get_edx_auth_failed():
+    """ Gets GeneratedCertificate model. """
```
anmrdz: As the [standard](https://www.python.org/dev/peps/pep-0257/#id16) describes: 

> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

So the (s) or conjugated verbs are not necessary.

eox_tenant/auth.py
```diff
@@ -0,0 +1,95 @@
...
+            # Run the default password hasher once to reduce the timing
+            # difference between an existing and a nonexistent user (#20760).
+            UserModel().set_password(password)
+        else:
+            if user.check_password(password) and self.user_can_authenticate_on_tenant(user, request):
+                return user
+
+    def user_can_authenticate_on_tenant(self, user, request):
+        """
+        Prevents users that signed up on a different tenant site to login in this site.
```
anmrdz:
```suggestion
        Prevent users that signed up on a different tenant site to login in this site.
```

eox_tenant/auth.py
morenol: The benefits of that is that we no longer will depend on ratelimitbackend

jfavellar90: Here I see two benefits: 1) not replicating the code from django, 2) avoid the dependency on ratelimitbackend. My approach here would be:

- Inherit directly from the edx-platform backend (I already tested the `super` calls and it should work as expected)
- Overwrite the `authenticate` method, add the line `self.request = request` and return the result of the `super` call
- Overwrite the `user_can_authenticate` method and inside it, call the `user_can_authenticate_on_tenant` method based on the feature flag EDNX_TENANT_AWARE_AUTH.

eox_tenant/edxapp_wrapper/auth.py
jfavellar90: ohohohohohoh

eox_tenant/auth.py
morenol: I made those changes (A little bit different the last item)

eox_tenant/auth.py
```diff
@@ -0,0 +1,82 @@
...
+
+        if not is_authorized:
+            loggable_id = user.id if user else "<unknown>"
+            if settings.FEATURES.get('EDNX_ENABLE_STRICT_LOGIN', False):
+                AUDIT_LOG.warning(
+                    u"User `%s` tried to login in site `%s`, but was denied permission based on the signup sources.",
+                    loggable_id,
+                    current_domain,
+                )
+                raise AuthFailedError(_('User not authorized to perform this action'))
```
morenol: @jfavellar90 If we raise this exception, I think that the ratelimit backends will not work as expected. Maybe we should return false here. In that case, I think that the ratelimit backend will work but the message to the login failure will be 
"Email or password is incorrect"

morenol: I mean, this line is not being called if we raise the exception,
https://github.com/eduNEXT/edunext-platform/blob/4dab3d5b4146f4b463d20681304109ca1086ed48/openedx/core/djangoapps/user_authn/views/login.py#L217

morenol: Which means that  LoginFailures.increment_lockout_counter(user) is not being called.

eox_tenant/auth.py
```diff
@@ -0,0 +1,82 @@
...
```
morenol: If we raise this exception the RateLimitBackend will not work as expected

@jfavellar90 

I  think we should return False instead

morenol: The "problem" with that is that the error message will be 'Email or password is incorrect.'  instead of 'User not authorized to perform this action'


https://github.com/eduNEXT/edunext-platform/blob/4dab3d5b4146f4b463d20681304109ca1086ed48/openedx/core/djangoapps/user_authn/views/login.py#L217

jfavellar90:
@morenol I suppose RateLimitBackend will not work properly because it's not counting a "failed login because of wrong tenant" to complete the rate limit. It will just handle the wrong user-password attempts. I think is not a big issue and there's a workaround for it, so we can keep the `AuthFailedError` exception

morenol:
The RateLimitBackend will not work because here the AuthFailedError exception is not being handled 

https://github.com/brutasse/django-ratelimit-backend/blob/b70540668c01d610bef7a18ef35a1068f06f2a8f/ratelimitbackend/backends.py#L38

If we return False, instead of the raising of that exception it works fine.

eox_tenant/edxapp_wrapper/auth.py
anmrdz:
```suggestion
    """ Return GeneratedCertificate model. """
```

morenol: That's a previous commit, I am going to squash

eox_tenant/auth.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        Override user_can_authenticate method with the logic to be tenant aware.
+        """
+        return self.user_can_authenticate_on_tenant(user)
+
+    def user_can_authenticate_on_tenant(self, user):
+        """
+        Prevent users that signed up on a different tenant site to login in this site.
+        """
+        if not settings.FEATURES.get("EDNX_TENANT_AWARE_AUTH", False):
```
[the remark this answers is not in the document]
morenol: Done

eox_tenant/auth.py
```diff
@@ -0,0 +1,82 @@
...
+AUDIT_LOG = logging.getLogger("audit")
+EdxAuthBackend = get_edx_auth_backend()
+UserModel = get_user_model()
+
+
+class TenantAwareAuthBackend(EdxAuthBackend):
+    """
+    Authentication Backend class which will check if the user has a signupsource in the requested site.
+    """
+    def authenticate(self, request, username=None, password=None, **kwargs):
```
jfavellar90: I think we can call the `authenticate` method like this:
```suggestion
    def authenticate(self, request, **kwargs):
```

eox_tenant/auth.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        Override default method to store the request which is being used by the
+        user_can_authenticate_on_tenant method.
+        """
+        self.request = request  # pylint: disable=attribute-defined-outside-init
+        return super(TenantAwareAuthBackend, self).authenticate(
+            request=request, username=username, password=password, **kwargs
```
jfavellar90: According to the change above, this line turns into:
```suggestion
            request=request, **kwargs
```
like RateLimit backend call the parent in [here](https://github.com/brutasse/django-ratelimit-backend/blob/1.1.1/ratelimitbackend/backends.py#L38)

eox_tenant/settings/aws.py
```diff
@@ -43,3 +49,8 @@ def plugin_settings(settings):  # pylint: disable=function-redefined
             'eox_tenant.middleware.AvailableScreenMiddleware',
             'eox_tenant.middleware.MicrositeCrossBrandingFilterMiddleware',
         ]
+
+        if EDX_AUTH_BACKEND in settings.AUTHENTICATION_BACKENDS:
```
jfavellar90: I wonder if we should replace the EDX_AUTH_BACKEND in the same exact position (list index) on AUTHENTICATION_BACKENDS, and if the EDX_AUTH_BACKEND is not in the AUTHENTICATION_BACKENDS list, just append it

eox_tenant/auth.py
```diff
@@ -0,0 +1,82 @@
...
+class TenantAwareAuthBackend(EdxAuthBackend):
+    """
+    Authentication Backend class which will check if the user has a signupsource in the requested site.
+    """
+    def authenticate(self, request, username=None, password=None, **kwargs):
+        """
+        Override default method to store the request which is being used by the
+        user_can_authenticate_on_tenant method.
+        """
+        self.request = request  # pylint: disable=attribute-defined-outside-init
```
jfavellar90: just leave a comment here indicating we are doing this to use the request in a method the variable is not accessible

jfavellar90:
@morenol my definitive last comments

eox_tenant/auth.py
morenol: I added that to the docstring of the method, should I add that here too?

eox_tenant/settings/aws.py
morenol: I am ok with replace in the same exact position, but I am against append that backend if the EDX_AUTH_BACKEND is not in the AUTHENTICATION_BACKENDS .  

If an openedx instance use CAS, it would be undesired to append that backend.

eox_tenant/settings/aws.py
```diff
@@ -43,3 +50,5 @@ def plugin_settings(settings):  # pylint: disable=function-redefined
             'eox_tenant.middleware.AvailableScreenMiddleware',
             'eox_tenant.middleware.MicrositeCrossBrandingFilterMiddleware',
         ]
+
+        settings.AUTHENTICATION_BACKENDS = [EOX_TENANT_AUTH_BACKEND if (backend == EDX_AUTH_BACKEND) else backend for backend in settings.AUTHENTICATION_BACKENDS]  # pylint: disable=line-too-long
```
morenol: @jfavellar90 Not sure if there is a more clean way to do this.

jfavellar90: @morenol I think it will work OK. And I agree with the fact of not appending the backend. Eox-tenant should only force the overwriting of the default auth backend, which is the one defined in EDX_AUTH_BACKEND

eox_tenant/settings/aws.py
jfavellar90: @morenol you're right, if the expected default backend is not present, don't do anything

morenol:
I am going to add tests

PR #43, opened 2019-08-21
Adding MicrositeMiddleware for studio app JFI-4

cocococosti:


jfavellar90:
@cocococosti I'm wondering if the order matters here (middleware execution). I would say NO, @morenol @anmrdz do you have any concern?

cocococosti:
I'm not 100% sure but I don't think it matters. I tested this in dev and the settings were configured correctly.

PR #44, opened 2019-08-22
Fixing tests by moving import.

jfavellar90:
Fixing tests by moving import. For some reason, the tests were failing randomly. According to my investigation we apparently inserted this problem in HA-3 (https://circleci.com/gh/eduNEXT/eox-tenant/602) and from that moment we get the tests failure sometimes in python 2.7. I ran the tests with this fix 4 consecutive times with no failures. Strange thing here: I was no able to replicate the problem running the tests in dev.

**Reason of the change**
From the traceback you can check in https://circleci.com/gh/eduNEXT/eox-tenant/602 there were nested calls of **import_module** method and that was causing a strange behavior

I also changed the Makefile to use the same fail-under value for local coverage tests

@felipemontoya 
@morenol 
@anmrdz 
@cocococosti

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ requirements: ## install environment requirements
 
 run-test: clean ## Run test suite.
 	coverage run --source="." manage.py test
-	coverage report --fail-under=40
+	coverage report --fail-under=61
 
 run-quality-test: clean ## Run quality test.
 	pycodestyle ./eox_tenant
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -19,7 +19,7 @@ requirements: ## install environment requirements
 
 run-test: clean ## Run test suite.
 	coverage run --source="." manage.py test
-	coverage report --fail-under=40
+	coverage report --fail-under=51
```
morenol: Can we increase this to 61?

jfavellar90: Done

PR #47, opened 2019-08-30
Ignore case in signup source authenticaticon backend

morenol:
This backend should ignore case. Otherwise we would not be able to authenticate in

example.com if we have a signupsource.site like 'Example.com'


cocococosti:
btw fix this typo please :) "should have **been** denied based on the signup sources." https://github.com/eduNEXT/eox-tenant/blob/0f4eec85ad6f98f864bce4d3c81408bb81865dfd/eox_tenant/auth.py#L78

PR #48, opened 2019-09-03
Jhony/fixes eox tenant

jfavellar90:
Solving eox_tenant issues:

- Enabling users with specific permission to login to any tenant. The permission display name is **Can login to all eox-tenant Tenants** under the app labell **auth**

- Getting current request in custom auth backend using theming helpers method

@felipemontoya 
@morenol 
@anmrdz 
@cocococosti

### DIFF
--- a/eox_tenant/auth.py
+++ b/eox_tenant/auth.py
@@ -11,41 +11,47 @@ from django.contrib.auth import get_user_model
 from django.utils.translation import ugettext as _
 
 from eox_tenant.edxapp_wrapper.auth import get_edx_auth_backend, get_edx_auth_failed
+from eox_tenant.edxapp_wrapper.theming_helpers import get_theming_helpers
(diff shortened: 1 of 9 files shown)


eox_tenant/auth.py
```diff
@@ -11,41 +11,47 @@
...
     def user_can_authenticate_on_tenant(self, user):
         """
         Prevent users that signed up on a different tenant site to login in this site.
         """
-        current_domain = self.request.META.get("HTTP_HOST")
+        # Check if the user has a permission to login to all tenants
+        if user.has_perm(LOGIN_ALL_TENANTS_PERMISSION_NAME):
+            return True
+
+        request = theming_helpers.get_current_request()
```
morenol: Can this method be called outside the scope of a request? @jfavellar90

jfavellar90: @morenol good point, I've shielded that case (https://pythonhosted.org/django-crum/#get-current-request)

PR #49, opened 2019-09-05
[WIP] FFI-3 Trying to change course enrollment manager to filter enrollments by org.

cocococosti:
I haven't been able to test any of this because I keep getting `No module named common.djangoapps.student` error.

Not sure of the correct way to import modules from edunext-platform I just followed the example from the rest of the code.

I appreciate any help :)

### DIFF
--- a/eox_tenant/apps.py
+++ b/eox_tenant/apps.py
@@ -4,6 +4,7 @@ File configuration for eox-tenant.
 from django.apps import AppConfig
 from django.conf import settings as base_settings
 
+
 class EdunextOpenedxExtensionsTenantConfig(AppConfig):
     """
     App configuration
@@ -53,7 +54,15 @@ class EdunextOpenedxExtensionsTenantConfig(AppConfig):
         load_permissions()
 
         if base_settings.EOX_TENANT_CHANGE_ENROLLMENT_MANAGER:
+            # This code overrides the default manager for CourseEnrollment model
             from eox_tenant.edxapp_wrapper.enrollments import get_enrollments_model
             from eox_tenant.models import EdnxCourseEnrollmentManager
+
+            # Inspired from the three lines in
+            # https://github.com/django/django/blob/1.11.23/django/db/models/base.py#L364
+            manager = EdnxCourseEnrollmentManager()
+            manager.auto_created = True
             enrollments = get_enrollments_model()
-            enrollments.objects = EdnxCourseEnrollmentManager()
+            # Before adding the new manager, remove already loaded managers
+            enrollments._meta.local_managers = []  # pylint: disable=protected-access
+            enrollments.add_to_class('objects', manager)
--- a/eox_tenant/apps.py
+++ b/eox_tenant/apps.py
@@ -2,7 +2,7 @@
 File configuration for eox-tenant.
 """
 from django.apps import AppConfig
-
+from django.conf import settings as base_settings
 
 class EdunextOpenedxExtensionsTenantConfig(AppConfig):
     """
@@ -51,3 +51,9 @@ class EdunextOpenedxExtensionsTenantConfig(AppConfig):
         """
         from eox_tenant.permissions import load_permissions
         load_permissions()
+
+        if base_settings.EOX_TENANT_CHANGE_ENROLLMENT_MANAGER:
+            from eox_tenant.edxapp_wrapper.enrollments import get_enrollments_model
+            from eox_tenant.models import EdnxCourseEnrollmentManager
+            enrollments = get_enrollments_model()
+            enrollments.objects = EdnxCourseEnrollmentManager()
--- a/eox_tenant/edxapp_wrapper/backends/enrollments_h_v1.py
+++ b/eox_tenant/edxapp_wrapper/backends/enrollments_h_v1.py
@@ -1,0 +1,11 @@
+""" backend """
+from student import models  # pylint: disable=import-error
+
+
+def get_enrollments_model():
+    """ backend function """
+    return models.CourseEnrollment
+
+def get_enrollments_model_manager():
+    """ backend function """
+    return models.CourseEnrollmentManager
--- a/eox_tenant/edxapp_wrapper/backends/enrollments_h_v1.py
+++ b/eox_tenant/edxapp_wrapper/backends/enrollments_h_v1.py
@@ -6,6 +6,7 @@ def get_enrollments_model():
     """ backend function """
     return models.CourseEnrollment
 
+
 def get_enrollments_model_manager():
     """ backend function """
     return models.CourseEnrollmentManager
--- a/eox_tenant/edxapp_wrapper/enrollments.py
+++ b/eox_tenant/edxapp_wrapper/enrollments.py
@@ -9,6 +9,7 @@ def get_enrollments_model(*args, **kwargs):
     backend = import_module(backend_function)
     return backend.get_enrollments_model(*args, **kwargs)
 
+
 def get_enrollments_model_manager(*args, **kwargs):
     """ Get BaseMicrositeBackend. """
     backend_function = settings.EOX_TENANT_ENROLLMENTS_BACKEND
--- a/eox_tenant/edxapp_wrapper/enrollments.py
+++ b/eox_tenant/edxapp_wrapper/enrollments.py
@@ -1,0 +1,16 @@
+""" Backend abstraction. """
+from importlib import import_module
+from django.conf import settings
+
+
+def get_enrollments_model(*args, **kwargs):
+    """ Get BaseMicrositeBackend. """
+    backend_function = settings.EOX_TENANT_ENROLLMENTS_BACKEND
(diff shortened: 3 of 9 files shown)


eox_tenant/edxapp_wrapper/backends/enrollments_h_v1.py
```diff
@@ -0,0 +1,11 @@
+""" backend """
+from common.djangoapps.student import models  # pylint: disable=import-error
```
morenol:
```suggestion
from student import models  # pylint: disable=import-error
```
This solves the import problem.

eox_tenant/models.py
```diff
@@ -177,3 +180,27 @@ class Meta:
...
         """
         app_label = "eox_tenant"
+
+class EdnxCourseEnrollmentManager(CourseEnrollmentManager):
+    """
+    Custom manager for CourseEnrollment with Table-level filter methods.
+    """
+    def get_queryset(self):
+
+        orgs_to_include = get_microsite().get_value('course_org_filter')
```
morenol: Use configuration helpers

eox_tenant/models.py
```diff
@@ -177,3 +180,27 @@ class Meta:
...
+
+        # Make sure we have a list
+        if orgs_to_include and not isinstance(orgs_to_include, list):
+            orgs_to_include = [orgs_to_include]
+
+        if not orgs_to_include:
+            # Making orgs_to_include an empty iterable
+            orgs_to_include = []
```
morenol: I think that in this case we have to use something like:
`orgs_to_exclude = configuration_helpers.get_all_orgs()
   return super(CourseEnrollmentManager, self).get_queryset().exclude(course_id__org__in=orgs_to_exclude)

`

eox_tenant/models.py
```diff
@@ -177,3 +182,13 @@ class Meta:
...
         """
         app_label = "eox_tenant"
+
+class EdnxCourseEnrollmentManager(CourseEnrollmentManager):
+    """
+    Custom manager for CourseEnrollment with Table-level filter methods.
+    """
+    def get_queryset(self):
+
+        orgs_to_exclude = configuration_helpers.get_all_orgs()
```
jfavellar90: why do you call the get_all_orgs method instead of getting the orgs of the current microsite?

eox_tenant/apps.py
```diff
@@ -51,3 +51,9 @@ def ready(self):
         """
         from eox_tenant.permissions import load_permissions
         load_permissions()
+
+        if base_settings.EOX_TENANT_CHANGE_ENROLLMENT_MANAGER:
+            from eox_tenant.edxapp_wrapper.enrollments import get_enrollments
+            from eox_tenant.models import EdnxCourseEnrollmentManager
+            enrollments = get_enrollments()
+            enrollments.objects = EdnxCourseEnrollmentManager()
```
jfavellar90: does it work as we expect?

eox_tenant/edxapp_wrapper/backends/enrollments_h_v1.py
```diff
@@ -0,0 +1,11 @@
...
+from student import models  # pylint: disable=import-error
+
+
+def get_enrollments():
+    """ backend function """
+    return models.CourseEnrollment
+
+def get_enrollments_manager():
+    """ backend function """
+    return models.CourseEnrollmentManager
```
jfavellar90: Add new line at the end of the file

eox_tenant/edxapp_wrapper/enrollments.py
```diff
@@ -0,0 +1,16 @@
+""" Backend abstraction. """
+from importlib import import_module
+from django.conf import settings
+
+
+def get_enrollments(*args, **kwargs):
```
jfavellar90: I would call this method get_enrollments_model

eox_tenant/edxapp_wrapper/backends/enrollments_h_v1.py
```diff
@@ -0,0 +1,11 @@
+""" backend """
+from student import models  # pylint: disable=import-error
+
+
+def get_enrollments():
```
jfavellar90: I would call this method get_enrollments_model

eox_tenant/edxapp_wrapper/enrollments.py
```diff
@@ -0,0 +1,16 @@
...
+    """ Get BaseMicrositeBackend. """
+    backend_function = settings.EOX_TENANT_ENROLLMENTS_BACKEND
+    backend = import_module(backend_function)
+    return backend.get_enrollments(*args, **kwargs)
+
+def get_enrollments_manager(*args, **kwargs):
```
jfavellar90: I would call this method get_enrollments_model_manager

eox_tenant/edxapp_wrapper/backends/enrollments_h_v1.py
```diff
@@ -0,0 +1,11 @@
...
+    """ backend function """
+    return models.CourseEnrollment
+
+def get_enrollments_manager():
```
jfavellar90: I would call this method get_enrollments_model_manager

eox_tenant/settings/common.py
```diff
@@ -51,6 +51,8 @@ def plugin_settings(settings):
     settings.UTILS_MODULE_BACKEND = 'eox_tenant.edxapp_wrapper.backends.util_h_v1'
     settings.CHANGE_DOMAIN_DEFAULT_SITE_NAME = "stage.edunext.co"
     settings.EOX_TENANT_LOAD_PERMISSIONS = True
+    settings.EOX_TENANT_ENROLLMENTS_BACKEND = 'eox_tenant.edxapp_wrapper.backends.enrollments_h_v1'
```
jfavellar90: Can we override these variables via aws.py?

eox_tenant/apps.py
jfavellar90: I mean, this is doing the filtering as we expect? If the answer is yes, that's supercool

cocococosti: It's not working :( I had to do this `enrollments._meta.default_manager_name = EdnxCourseEnrollmentManager()` to make the code to run at least but the results are not being filtered as expected and also it's not a good practice.

eox_tenant/edxapp_wrapper/backends/enrollments_h_v1.py
```diff
@@ -0,0 +1,11 @@
+""" backend """
+from student import models  # pylint: disable=import-error
+
+
+def get_enrollments_model():
+    """ backend function """
```
anmrdz: @cocococosti please improve the description in this docstring.

eox_tenant/edxapp_wrapper/backends/enrollments_h_v1.py
```diff
@@ -0,0 +1,11 @@
...
+    return models.CourseEnrollment
+
+def get_enrollments_model_manager():
+    """ backend function """
```
anmrdz: @cocococosti same comment as above.

eox_tenant/edxapp_wrapper/enrollments.py
```diff
@@ -0,0 +1,16 @@
+""" Backend abstraction. """
```
anmrdz: @cocococosti I think is better for clarity and future code maintenance that this docstring is more specific as to what backend is referring to.

eox_tenant/models.py
```diff
@@ -177,3 +182,24 @@ class Meta:
         Model meta class.
         """
         app_label = "eox_tenant"
+
+class EdnxCourseEnrollmentManager(CourseEnrollmentManager):
+    """
+    Custom manager for CourseEnrollment with Table-level filter methods.
+    """
+    def get_queryset(self):
+
```
anmrdz: @cocococosti Please add the missing docstring here.

eox_tenant/settings/common.py
anmrdz: @cocococosti @jfavellar90 I don't know if this applies here but I think this file should point to 
```suggestion
    settings.EOX_TENANT_ENROLLMENTS_BACKEND = 'eox_tenant.edxapp_wrapper.backends.enrollments_i_v1'
```
and the corresponding file should be renamed to that version too, Does that make sense here?

cocococosti: I named the files following the example of the other ones, the i in enrollments_i_v1 is for ironwood?

eox_tenant/edxapp_wrapper/backends/enrollments_h_v1.py
cocococosti: If this PR take us anywhere, doesn't work for me but Jhony is trying to fix it, I'll write better docstrings

eox_tenant/settings/common.py
morenol: Yes, that is the meaning

PR #51, opened 2019-09-15
Changing the change_domain command to affect the ecommerce urls as well

jfavellar90:
Changing the change_domain command to affect the ecommerce urls as well

@morenol 
@cocococosti


jfavellar90:
Tests OK so I proceed to merge

PR #54, opened 2019-10-04
Changes to fix edxapp tests

jfavellar90:
Try-catching the custom permissions loading if the django tables don't exist yet. This will fix the edxapp tests run

@felipemontoya 
@morenol 
@cocococosti


morenol:
Tests passed on LMS and CMS

PR #58, opened 2019-12-02
tenancy for async tasks

morenol:
Handle celery signals (celery.signals.before_task_publish and celery.signals.task_prerun) to change the settings of the async process accordingly to the tenant configurations.

### DIFF
--- a/eox_tenant/signals.py
+++ b/eox_tenant/signals.py
@@ -23,13 +23,13 @@ from __future__ import print_function, unicode_literals
 import logging
 import os
 from datetime import datetime
-
 import six
 
 from django.apps.config import AppConfig
 from django.conf import settings as base_settings
 
 from eox_tenant.backends.database import TenantConfigCompatibleMicrositeBackend
+from eox_tenant.async_utils import AsyncTaskHandler
 
 LOG = logging.getLogger(__name__)
 
@@ -63,6 +63,13 @@ def _update_settings(domain):
             continue
         setattr(base_settings, key, value)
 
+    # Some django apps need to be reinitialized
+    try:
+        if base_settings.EDNX_TENANT_INSTALLED_APPS:
+            _repopulate_apps(base_settings.EDNX_TENANT_INSTALLED_APPS)
+    except AttributeError:
+        pass
+
 
 def _repopulate_apps(apps):
     """
@@ -76,6 +83,20 @@ def _repopulate_apps(apps):
     for entry in apps:
         app_config = AppConfig.create(entry)
         app_config.ready()
+
+
+def can_keep_settings(domain):
+    """
+    Find what we need to do about the current setting and the incoming request.domain
+    and Reset settings if needed.
+    """
+    must_reset, can_keep = _analyze_current_settings(domain)
+
+    if _ttl_reached() or must_reset:  # Perform the reset
+        _perform_reset()
+        can_keep = False
+
+    return can_keep
 
 
 def _analyze_current_settings(domain):
@@ -174,29 +195,11 @@ def start_tenant(sender, environ, **kwargs):  # pylint: disable=unused-argument
 
     domain = http_host.split(':')[0]
 
-    # Find what we need to do about the current setting and the incoming request.domain
-    must_reset, can_keep = _analyze_current_settings(domain)
-
-    # Reset minimum once every so often
-    must_reset = _ttl_reached() or must_reset
-
-    # Perform the reset
-    if must_reset:
-        _perform_reset()
-        can_keep = False
-
-    if can_keep:
+    if can_keep_settings(domain):
         return
 
     # Do the update
     _update_settings(domain)
-
-    # Some django apps need to be reinitialized
-    try:
-        if base_settings.EDNX_TENANT_INSTALLED_APPS:
-            _repopulate_apps(base_settings.EDNX_TENANT_INSTALLED_APPS)
-    except AttributeError:
-        pass
 
 
 def finish_tenant(sender, **kwargs):  # pylint: disable=unused-argument
@@ -217,3 +220,39 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
     Signal: django.core.signals.got_request_exception
     """
     pass
+
+
+def tenant_context_addition(sender, body, headers, *args, **kwargs):  # pylint: disable=unused-argument
+    """
+    Receiver used to add context to the async process about the tenant associated to the task that need to be done.
+
+    Dispatched before a celery task is published. Note that this is executed in the process sending the task.
+    See:
+        https://celery.readthedocs.io/en/latest/userguide/signals.html#before-task-publish
+    """
+    get_host_func = AsyncTaskHandler().get_host_from_task(sender)
+    headers['eox_tenant_sender'] = get_host_func(body)
+
+
+def start_async_tenant(sender, *args, **kwargs):  # pylint: disable=unused-argument
+    """
+    Receiver that runs on the async process to update the settings accordingly to the tenant.
+    Dispatched before a task is executed.
+    See:
+       https://celery.readthedocs.io/en/latest/userguide/signals.html#task-prerun
+    """
+    headers = sender.request.get('headers') or {}
+    http_host = headers.get('eox_tenant_sender')
+
+    if not http_host:  # Reset settings in case of no tenant.
+        LOG.warning("Could not find the host information for eox_tenant.signals. ")
+        _perform_reset()
+        return
+
+    domain = http_host
+
+    if can_keep_settings(domain):
+        return
+
+    # Do the update
+    _update_settings(domain)
--- a/eox_tenant/test/test_signals.py
+++ b/eox_tenant/test/test_signals.py
@@ -6,6 +6,7 @@ from __future__ import print_function, unicode_literals
 
 from datetime import datetime, timedelta
 
+from django.contrib.sites.models import Site
 from django.test import TestCase
 from mock import MagicMock, patch
 
@@ -15,6 +16,8 @@ from eox_tenant.signals import (
     _update_settings,
     _analyze_current_settings,
     _repopulate_apps,
+    tenant_context_addition,
+    start_async_tenant,
 )
 
 
@@ -162,13 +165,18 @@ class StartTenantSignalTest(TestCase):
 
         _repopulate_mock.assert_not_called()
 
+    @patch('eox_tenant.signals._get_tenant_config')
     @patch('eox_tenant.signals._repopulate_apps')
-    def test__repopulate_apps_called(self, _repopulate_mock):
+    def test__repopulate_apps_called(self, _repopulate_mock, _get_config_mock):
         """
         A tenant that defines the EDNX_TENANT_INSTALLED_APPS calls the function
         """
         environ = MagicMock()
         environ.get.return_value = 'tenant.com'
+        config = {
+            "EDNX_USE_SIGNAL": True,
+        }
+        _get_config_mock.return_value = config, "tenant-key"
 
         with self.settings(SERVICE_VARIANT='lms', EDNX_TENANT_INSTALLED_APPS=['fake_app']):
             start_tenant(None, environ)
@@ -283,3 +291,101 @@ class SettingsOverridesTest(TestCase):
 
         self.assertEqual(settings.TEST_DICT_OVERRIDE_TEST.get("key1"), "Some Value")
         self.assertEqual(settings.TEST_DICT_OVERRIDE_TEST.get("key2"), "My value")
+
+
+class CeleryReceiverCLISyncTests(TestCase):
+    """
+    Testing the celery signals generated outside of a request in a sync process.
+    """
+
+    def setUp(self):
+        """ setup """
+        for number in range(3):
+            Site.objects.create(  # pylint: disable=no-member
+                domain="tenant{number}.com".format(number=number),
+                name="tenant{number}.com".format(number=number)
+            )
+
+    def test_unknown_task(self):
+        """
+        When the task is unknown the hostname should be None.
+        """
+        headers = {}
+        body = {}
+        tenant_context_addition("uknown_task", headers=headers, body=body)
+
+        self.assertIsNone(headers['eox_tenant_sender'])
+
+    @patch('eox_tenant.async_utils.AsyncTaskHandler.get_host_from_siteid')
+    def test_send_recurring_nudge(self, _get_host_mock):
+        """
+        When the task is called the method get_host_from_siteid should be called.
+        """
+        headers = {}
+        body = {}
+
+        with self.settings():
+            tenant_context_addition(
+                "openedx.core.djangoapps.schedules.tasks.ScheduleRecurringNudge",
+                headers=headers,
+                body=body,
+            )
+
+        _get_host_mock.assert_called()
+
+
+class CeleryReceiverSyncTests(TestCase):
+    """
+    Testing the celery receivers that run in a sync process
+    """
+
+    def setUp(self):
+        """ setup """
+        for number in range(3):
+            Site.objects.create(  # pylint: disable=no-member
+                domain="tenant{number}.com".format(number=number),
+                name="tenant{number}.com".format(number=number)
+            )
+
+    def test_sync_process_with_tenant(self):
+        """
+        When the task is called from a sync proccess it is used the value of the 'host' key in the request dict.
+        """
+        headers = {}
+        body = {}
+        with self.settings(EDNX_TENANT_DOMAIN="some.tenant.com"):
+            tenant_context_addition("some.task", headers=headers, body=body)
+
+        self.assertEqual(headers['eox_tenant_sender'], 'some.tenant.com')
+
+
+class CeleryReceiverAsyncTests(TestCase):
+    """
+    Testing the signal handling in the async process.
+    """
+
+    @patch('eox_tenant.signals._update_settings')
+    def test_start_async_tenant(self, _update_mock):
+        """
+        Test function used to handle signal in the async process
+        """
+        sender = MagicMock()
+        headers = {"eox_tenant_sender": "some.tenant.com"}
+        sender.request = {"headers": headers}
+        start_async_tenant(sender)
+
+        _update_mock.assert_called_with("some.tenant.com")
+
+    @patch('eox_tenant.signals._perform_reset')
+    @patch('eox_tenant.signals._update_settings')
+    def test_start_async_tenant_with_null_header(self, _update_mock, _reset_mock):
+        """
+        Test function used to handle signal in the async process.
+        """
+        sender = MagicMock()
+        headers = None
+        sender.request = {"headers": headers}
+        start_async_tenant(sender)
+
+        _reset_mock.assert_called()
+        _update_mock.assert_not_called()
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
           command: |
             . venv/bin/activate
             coverage run --source="." manage.py test
-            coverage report --fail-under=61 -m
+            coverage report --fail-under=71 -m
             pycodestyle ./eox_tenant
             pylint ./eox_tenant --rcfile=./setup.cfg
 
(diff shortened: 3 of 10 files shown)


eox_tenant/signals.py
```diff
@@ -45,6 +45,12 @@ def _update_settings(domain):
 
     if not config.get("EDNX_USE_SIGNAL"):
         LOG.info("Site %s, does not use eox_tenant signals", domain)
+        # Some django apps need to be reinitialized
```
felipemontoya: We could have this block and the next block that does the same refactored into a function.

However EDNX_TENANT_INSTALLED_APPS will never exist for sites that do not have EDNX_USE_SIGNAL unless we set that in the settings.common module.
I would also suggest that if a site does not have ednx_use_signal we should ignore the repopulate apps procedure. We should not expect a site to work properly in any case if it does not use signals

eox_tenant/signals.py
```diff
@@ -174,30 +204,14 @@ def start_tenant(sender, environ, **kwargs):  # pylint: disable=unused-argument
...
-        can_keep = False
+    can_keep = reset_or_keep_settings(domain)
 
     if can_keep:
         return
 
     # Do the update
     _update_settings(domain)
 
-    # Some django apps need to be reinitialized
```
felipemontoya: +1 about moving this into the _update

eox_tenant/signals.py
```diff
@@ -217,3 +231,90 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+    body = kwargs.get('body')
+    try:
+        request_info = body.get('args')[1].get('request_info')
+        headers['eox_tenant_sender'] = request_info.get('host') or None
+    except AttributeError:
+        # This is a async taks generated outside of the scope of a request i.e. CLI task
+        get_host_func = AsyncTaskHandler().get_host_from_task(sender)
+        headers['eox_tenant_sender'] = get_host_func(body)
+
+    headers['X-Forwarded-For'] = 'CRITICAL_PIECE_OF_INFO'
```
felipemontoya: this line was part of the POC can be removed with no issues

eox_tenant/signals.py
```diff
@@ -217,3 +231,90 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+        Get default tenant.
+        """
+        def get_host(_body):
+            """
+            Get host using default site id.
+            """
+            siteid = base_settings.SITE_ID
+            site = Site.objects.get(id=siteid)
+            return site.name
+        return get_host
```
felipemontoya: I believe this is the place where we should be logging the fact that we did not find the tenant and we need to raise the alarms.

eox_tenant/signals.py
morenol: I added this block of code in this part due to a failing test (test__repopulate_apps_called)

 I am going to fix that!

andrey-canon:
@morenol A description would be nice

eox_tenant/signals.py
```diff
@@ -217,3 +225,97 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+    }
+
+    sender = None
+
+    def get_host_from_task(self, sender):
+        """
+        Method used to get the function used to get the tenant domain
+        accordingly to the type of task.
+        """
+        self.sender = sender
```
andrey-canon: @morenol  self.sender is not used, you could remove it

eox_tenant/signals.py
```diff
@@ -217,3 +225,97 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+    # Do the update
+    _update_settings(domain)
+
+
+class AsyncTaskHandler(object):
+    """
+    Handler used to get the tenant of an async task outside of the scope of a request.
+    """
+    TASKS = {
+        "openedx.core.djangoapps.schedules.tasks.ScheduleRecurringNudge": 'get_host_from_siteid'
```
andrey-canon: @morenol  why only this kind of task?

eox_tenant/signals.py
```diff
@@ -217,3 +225,97 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+        """
+        Get the tenant domain name using its Site id
+        """
+        def get_host(body):
+            """
+            Get host from siteid.
+            """
+            siteid = body.get('args')[0]
+            site = Site.objects.get(id=siteid)
+            return site.name
```
andrey-canon: @morenol  add blank line after this line

eox_tenant/signals.py
```diff
@@ -217,3 +225,97 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+    def get_host_from_siteid(self):
+        """
+        Get the tenant domain name using its Site id
+        """
+        def get_host(body):
+            """
+            Get host from siteid.
+            """
+            siteid = body.get('args')[0]
+            site = Site.objects.get(id=siteid)
```
andrey-canon: this raises a `DoesNotExist ` exception, we should handle it

eox_tenant/signals.py
```diff
@@ -217,3 +225,97 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+
+    def get_host_from_siteid(self):
+        """
+        Get the tenant domain name using its Site id
+        """
+        def get_host(body):
+            """
+            Get host from siteid.
+            """
+            siteid = body.get('args')[0]
```
andrey-canon: handle possible exceptions

eox_tenant/signals.py
```diff
@@ -217,3 +225,97 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+
+    # Do the update
+    _update_settings(domain)
+
+
+class AsyncTaskHandler(object):
+    """
+    Handler used to get the tenant of an async task outside of the scope of a request.
+    """
+    TASKS = {
```
andrey-canon: @morenol  I think this should be a plugin setting

eox_tenant/test/test_signals.py
```diff
@@ -283,3 +292,95 @@ def test_udpate_settings_with_existing_dict(self, _get_config_mock):
 
         self.assertEqual(settings.TEST_DICT_OVERRIDE_TEST.get("key1"), "Some Value")
         self.assertEqual(settings.TEST_DICT_OVERRIDE_TEST.get("key2"), "My value")
+
+
+class CelerySignalCLISyncTests(TestCase):
```
andrey-canon: @morenol  we are not testing the signals we are testing the receivers, so imo this should be called different

eox_tenant/test/test_signals.py
```diff
@@ -283,3 +292,95 @@ def test_udpate_settings_with_existing_dict(self, _get_config_mock):
...
+    """
+    def setUp(self):
+        """ setup """
+        for number in range(3):
+            Site.objects.create(  # pylint: disable=no-member
+                domain="tenant{number}.com".format(number=number),
+                name="tenant{number}.com".format(number=number)
+            )
+
+    def test_get_host_from_siteid(self):
```
andrey-canon: here you don't test a signal or receiver, so  imo this deserves an own testing class

eox_tenant/test/test_signals.py
```diff
@@ -283,3 +292,95 @@ def test_udpate_settings_with_existing_dict(self, _get_config_mock):
...
+        When the task is called the method get_host_from_siteid should be called.
+        """
+        headers = {}
+        with self.settings():
+            before_task_publish("openedx.core.djangoapps.schedules.tasks.ScheduleRecurringNudge", headers=headers)
+
+        _get_host_mock.assert_called()
+
+
+class CelerySignalSyncTests(TestCase):
```
andrey-canon: same comment as above

eox_tenant/test/test_signals.py
```diff
@@ -283,3 +292,95 @@ def test_udpate_settings_with_existing_dict(self, _get_config_mock):
...
+        siteid = 1
+        body['args'] = [siteid, {'request_info': {'host': 'some.tenant.com'}}]
+
+        with self.settings():
+            before_task_publish("some.task", headers=headers, body=body)
+
+        self.assertEqual(headers['eox_tenant_sender'], 'some.tenant.com')
+
+
+class CelerySignalAsyncTests(TestCase):
```
andrey-canon: same comment as above

eox_tenant/test/test_signals.py
```diff
@@ -283,3 +292,95 @@ def test_udpate_settings_with_existing_dict(self, _get_config_mock):
...
+            before_task_publish("uknown_task", headers=headers)
+
+        self.assertEqual('tenant1.com', headers['eox_tenant_sender'])
+
+    @patch('eox_tenant.signals.AsyncTaskHandler.get_host_from_siteid')
+    def test_send_recurring_nudge(self, _get_host_mock):
+        """
+        When the task is called the method get_host_from_siteid should be called.
+        """
+        headers = {}
```
andrey-canon: add blank line after this line

eox_tenant/test/test_signals.py
```diff
@@ -283,3 +292,95 @@ def test_udpate_settings_with_existing_dict(self, _get_config_mock):
...
+        hostname = func(body)
+        self.assertEqual(hostname, "tenant2.com")
+
+    def test_unknown_task(self):
+        """
+        When the task is unknown the hostname used should
+        be the same associated to the SITE_ID setting.
+        """
+        headers = {}
+        site = Site.objects.get(domain="tenant1.com")
```
andrey-canon: same comment as above

eox_tenant/test/test_signals.py
```diff
@@ -283,3 +292,95 @@ def test_udpate_settings_with_existing_dict(self, _get_config_mock):
...
+
+    def test_get_host_from_siteid(self):
+        """
+        Test method used to get host from siteid
+        """
+        func = AsyncTaskHandler().get_host_from_siteid()
+        site = Site.objects.get(domain="tenant2.com")
+        body = {}
+        body['args'] = (site.id, None, None)
+
```
andrey-canon: unnecessary blank line

eox_tenant/test/test_signals.py
```diff
@@ -283,3 +292,95 @@ def test_udpate_settings_with_existing_dict(self, _get_config_mock):
...
+        with self.settings():
+            before_task_publish("some.task", headers=headers, body=body)
+
+        self.assertEqual(headers['eox_tenant_sender'], 'some.tenant.com')
+
+
+class CelerySignalAsyncTests(TestCase):
+    """
+    Testing the signal handling in the async process.
+    """
```
andrey-canon: add blank line after this line

eox_tenant/test/test_signals.py
```diff
@@ -283,3 +292,95 @@ def test_udpate_settings_with_existing_dict(self, _get_config_mock):
...
+        with self.settings():
+            before_task_publish("openedx.core.djangoapps.schedules.tasks.ScheduleRecurringNudge", headers=headers)
+
+        _get_host_mock.assert_called()
+
+
+class CelerySignalSyncTests(TestCase):
+    """
+    Testing the celery signals generated from a sync process
+    """
```
andrey-canon: add blank line after this line

eox_tenant/signals.py
```diff
@@ -217,3 +225,97 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+        return
+
+    # Do the update
+    _update_settings(domain)
+
+
+class AsyncTaskHandler(object):
+    """
+    Handler used to get the tenant of an async task outside of the scope of a request.
+    """
```
andrey-canon: add blank line after this line

eox_tenant/signals.py
```diff
@@ -217,3 +225,97 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
     Signal: django.core.signals.got_request_exception
     """
     pass
+
+
+def before_task_publish(sender, *args, **kwargs):  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
felipemontoya: I agree with the name change. How do you like `tenant_context_addition` which is the object of this function

eox_tenant/signals.py
```diff
@@ -217,3 +225,97 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+    """
+    Dispatched before a celery task is published. Note that this is executed in the process sending the task.
```
felipemontoya: This should explain the reasoning for this function so that a human can understand the purpose of the function. Also, as the previous comment noted, this is the receiver of the signal.

eox_tenant/signals.py
morenol: It is used in the logging in the default case

eox_tenant/signals.py
morenol: This is the only CLI task that I know how to handle right now.

eox_tenant/signals.py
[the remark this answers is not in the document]
morenol: I changed that to `receiver_before_task_publish`, is it OK? or should I change that to other name?

eox_tenant/signals.py
```diff
@@ -217,3 +225,97 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+    See:
+        https://celery.readthedocs.io/en/latest/userguide/signals.html#before-task-publish
+    """
+    headers = kwargs.get('headers')
+    body = kwargs.get('body')
```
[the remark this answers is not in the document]
morenol: I am not sure which default I should add to this value.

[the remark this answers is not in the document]
andrey-canon:
```suggestion
    body = kwargs.get('body', {})
```

eox_tenant/signals.py
```diff
@@ -217,3 +225,97 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+       https://celery.readthedocs.io/en/latest/userguide/signals.html#task-prerun
+    """
+    headers = sender.request.get('headers', {})
+
+    http_host = headers.get('eox_tenant_sender', None)
+
+    if not http_host:
+        LOG.warning("Could not find the host information for eox_tenant.signals")
+
+    domain = http_host.split(':')[0]
```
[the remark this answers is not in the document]
morenol: I think that in the previous if, I should set that variable as the default http_host, what do you think?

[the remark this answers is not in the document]
andrey-canon: maybe something like this `http_host = headers.get('eox_tenant_sender',  default_host(body))` in line 255, you could delete the if statement  because you log that in line 307, but I don't know  if you can get the body here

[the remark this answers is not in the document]
andrey-canon: I think it's better to  set the default value inside the if statement @morenol  forget the previous message

morenol:
Files with name tasks.py that are in openedx:

./openedx/core/djangoapps/heartbeat/tasks.py
./openedx/core/djangoapps/service_status/tasks.py
./openedx/core/djangoapps/credit/tasks.py
./openedx/core/djangoapps/verified_track_content/tasks.py
./openedx/core/djangoapps/ccxcon/tasks.py
./openedx/core/djangoapps/credentials/tasks/v1/tasks.py
./openedx/core/djangoapps/programs/tasks/v1/tasks.py
./openedx/core/djangoapps/bookmarks/tasks.py
./openedx/core/djangoapps/content/block_structure/tasks.py
./openedx/core/djangoapps/content/course_overviews/tasks.py
./openedx/core/djangoapps/coursegraph/tasks.py
./openedx/core/djangoapps/schedules/tasks.py
./common/djangoapps/student/tasks.py
./common/djangoapps/entitlements/tasks.py
./common/djangoapps/third_party_auth/tasks.py
./cms/djangoapps/cms_user_tasks/tasks.py
./cms/djangoapps/contentstore/tasks.py
./lms/djangoapps/gating/tasks.py
./lms/djangoapps/grades/tasks.py
./lms/djangoapps/certificates/tasks.py
./lms/djangoapps/bulk_email/tasks.py
./lms/djangoapps/lti_provider/tasks.py
./lms/djangoapps/verify_student/tasks.py
./lms/djangoapps/ccx/tasks.py
./lms/djangoapps/email_marketing/tasks.py
./lms/djangoapps/instructor_task/tasks.py
./lms/djangoapps/discussion/tasks.py

eox_tenant/signals.py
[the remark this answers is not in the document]
felipemontoya: A function's name should give you an idea of what the function does, not give you an idea of where the function is called. I'd say, lets change it.

eox_tenant/signals.py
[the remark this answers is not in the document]
felipemontoya: There is no real need to set a default here. Without it, the function is going to fail anyways and the signal that sends this message always sends  a `headers` and a `body`. In fact probably the best thing here would be to make it explicit by making the `function(sender, *args, **kwargs)` signature match the one sent by the `before_task_publish` signal

eox_tenant/signals.py
```diff
@@ -217,3 +222,114 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+    headers = sender.request.get('headers') or {}
+    http_host = headers.get('eox_tenant_sender')
+    if not http_host:
+        LOG.warning("Could not find the host information for eox_tenant.signals. ")
+        siteid = base_settings.SITE_ID
+        site = Site.objects.get(id=siteid)
+        http_host = site.name
+
+    try:
+        domain = http_host.split(':')[0]
```
andrey-canon: you need this domain to execute the method `_update_settings` but `_update_settings`, after a lot of method calls, executes this line with the domain parameter https://github.com/eduNEXT/eox-tenant/blob/master/eox_tenant/models.py#L149, so why don't remove this or add this to the `get_host` methods that returns the domain

eox_tenant/signals.py
```diff
@@ -217,3 +222,114 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+
+    if reset_or_keep_settings(domain):
+        return
+
+    # Do the update
+    _update_settings(domain)
+
+
+class AsyncTaskHandler(object):
```
andrey-canon: we could move this to another  file, since this doesn't have any relationship with signals

eox_tenant/signals.py
```diff
@@ -217,3 +222,114 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+    See:
+       https://celery.readthedocs.io/en/latest/userguide/signals.html#task-prerun
+    """
+    headers = sender.request.get('headers') or {}
+    http_host = headers.get('eox_tenant_sender')
+    if not http_host:
+        LOG.warning("Could not find the host information for eox_tenant.signals. ")
+        siteid = base_settings.SITE_ID
+        site = Site.objects.get(id=siteid)
+        http_host = site.name
```
felipemontoya: Why is the default tenant that of the `base_settings.SITE_ID` ?
Sure, django requires a default Site ID, but that does not mean that we are better of setting the tenants to that when we can not find the correct setting.

I would say that instead we can use a bogus string (e.g `"<no_tenant_selected>"`) or None as the domain and let the reset_or_keep_settings function handle it.

This is my opinion at least, better to have no tenant activated than to select the incorrect tenant by default.

eox_tenant/signals.py
```diff
@@ -174,30 +197,12 @@ def start_tenant(sender, environ, **kwargs):  # pylint: disable=unused-argument
...
-    # Reset minimum once every so often
-    must_reset = _ttl_reached() or must_reset
-
-    # Perform the reset
-    if must_reset:
-        _perform_reset()
-        can_keep = False
-
-    if can_keep:
+    if reset_or_keep_settings(domain):
```
felipemontoya: Ironically the `reset_or_keep_settings` function returns a Boolean for an answer? instead of the string `reset` or `keep`. I would say the function should be called "can_keep_settings?" because then the True/False value makes more sense

eox_tenant/signals.py
```diff
@@ -217,3 +222,114 @@ def clear_tenant(sender, request, **kwargs):  # pylint: disable=unused-argument
...
+                return self.default_host()(body)
+            return domain
+
+        return get_host
+
+    def default_host(self):
+        """
+        Get default tenant.
+        """
+        LOG.warning(
```
felipemontoya: If this is the default way of getting the context, I do not think that such a warning is required just yet.

First we try to find if we are in a sync process that has already an active tenant. If so we use that tenant.
Then if no tenant is available we log the error and as much information as we can. E.g The traceback.

PR #59, opened 2019-12-02
Remove duplicated method

morenol:
This method is already defined in ox_tenant.backends.base.BaseMicrositeBackend


jfavellar90:
LGTM unless someone is expecting to find logs with the "Eox-tenant" text

felipemontoya:
@morenol if you are testing this locally and it runs then I'm happy about removing it.
We want to get all references of microsites out by juniper. However we need to test this in stage to make sure it will not interrupt operations for the saas platform.

PR #61, opened 2019-12-04
Remove edx microsite mentions

andrey-canon:
## Description
The work of this PR is based on we don't want to use the edx microsite_configuration app, since this was removed in this [PR](https://github.com/edx/edx-platform/pull/21380)
####  (First commit)Remove get_microsite backend
- Remove middlewares(issue JFI-4 right now cms doesn't use microsites in the code, so I didn't find a reason to set the microsite using the domain)
- Remove proxy(This overrode the settings using the microsite settings, but now we don't use the microsite settings )
- Remove tests

####  (Second commit)Remove BaseMicrositeTemplateBackend
- there was a recursive call in the file microsite_configuration_h_v1.py, so this class only was used directly on the test settings.

#### (Third commit) Remove EdunextCompatibleFilebasedMicrositeTemplateBackend.
- we don't need anymore  to override the MICROSITE_TEMPLATE_BACKEND setting.

#### (Fourth commit) Remove get_base_microsite_template_backend.
- The classes, that depended on this, were removed

#### (Fifth commit) Remove get_is_request_in_microsite.
- This was necessary in the file eox_tenant/backends/filebased.py, but it was removed.

#### (Sixth commit) Remove get_microsite_get_value.
- This was necessary in the file eox_tenant/backends/filebased.py, but it was removed.

#### (Seventh commit) Remove get_base_microsite_backend.
- Use AbstractBaseMicrositeBackend as default backend for  BaseMicrositeBackend

#### (Eighth commit) Remove unnecessary settings.
- Remove unnecessary settings.

#### (Nineth commit) Remove backend files and use a new model.
- Move all the logic from the backend folder to the TenantConfigCompatibleSiteConfigurationProxyModel
- Remove unnecessary files.
- Add middleware to do monkey patch to the SiteConfiguration model.
- Remove backend test and and model tests.
- Rename configuration helpers files.


morenol:
I think that the reason we had to use microsites middleware in studio is due to this feature: LC-V2

https://github.com/eduNEXT/edunext-platform/pull/305

andrey-canon:
> k that the reason we had to use microsites middleware in studio is due to this feature: L

@morenol I have found this document https://docs.google.com/document/d/1EO5aDX8CrsDRwdlqtu-VE5SduQmQ3xYLpShfOdhXYqQ/edit# but it says that there is nothing to test in studio... do you now how to test it ?

eox_tenant/backends/base.py
```diff
@@ -103,12 +100,7 @@ def clear(self):
         raise NotImplementedError()
 
 
-# This is the connection point between openedx and this module
-# We need to pass the class inheritance validation, and also stand alone for tests
-MB_INTERFACE_CONNECTION_BACKEND = get_base_microsite_backend()
-
-
-class BaseMicrositeBackend(MB_INTERFACE_CONNECTION_BACKEND):
+class BaseMicrositeBackend(AbstractBaseMicrositeBackend):
```
felipemontoya: Do we need this class anymore?

felipemontoya:
> > k that the reason we had to use microsites middleware in studio is due to this feature: L
> 
> @morenol I have found this document https://docs.google.com/document/d/1EO5aDX8CrsDRwdlqtu-VE5SduQmQ3xYLpShfOdhXYqQ/edit# but it says that there is nothing to test in studio... do you now how to test it ?

If we decided to support studio in eox-tenant then any usage of microsite there would become irrelevant. We should aim there instead of keeping some old feature just to not break anything

morenol:
> We should aim there instead of keeping some old feature just to not break anything

I agree that we have to remove this old feature. I just wanted to add that if we remove this feature, then we need to do the changes to do not break the language selector support in studio.

eox_tenant/edxapp_wrapper/backends/site_configuration_module_h_v1.py
```diff
@@ -0,0 +1,17 @@
+""" backend """
```
morenol: Can you change the name to site_configuration_module_i_v1.py?

eox_tenant/middleware.py
```diff
@@ -126,3 +92,16 @@ def process_request(self, request):
                     {'domain': domain, }
                 )
             )
+
+
+class MonkeyPatchMiddleware(object):
```
jfavellar90: I don't know so much about django monkey patching best practices but what about monkeypatching the SiteConfiguration model after Django load? does it pose some technical problem?

andrey-canon: @jfavellar90 I don't know if monkey patch has best practices, since the most of people recommends don't use that,anyway  another good place to do this is in the apps.py  in the method ready, the problem, in that point, is that the settings has not been loaded yet, as we use the setting module to load all the backends, we could not get the backend in the standard way, another reason to make this in the middleware was this article https://opensourcehacker.com/2010/08/12/applying-monkey-patches-in-django-middleware-layer/

eox_tenant/middleware.py
```diff
@@ -11,58 +11,26 @@
...
-from eox_tenant.edxapp_wrapper.configuration_helpers import get_configuration_helpers
 from eox_tenant.edxapp_wrapper.edxmako_module import get_edxmako_module
-from eox_tenant.edxapp_wrapper.get_microsite_configuration import get_microsite
+from eox_tenant.edxapp_wrapper.site_configuration_module import get_configuration_helpers, get_site_configuration_models
 from eox_tenant.edxapp_wrapper.theming_helpers import get_theming_helpers
+from eox_tenant.monkey_patch.monkey_patch_proxys import TenantSiteConfigProxy
 
-microsite = get_microsite()  # pylint: disable=invalid-name
 configuration_helper = get_configuration_helpers()  # pylint: disable=invalid-name
+SiteConfigurationModels = get_site_configuration_models()
```
morenol: I think that this is not being used in this file

eox_tenant/monkey_patch/monkey_patch_proxys.py
```diff
@@ -0,0 +1,160 @@
...
+    This allows to add or override methods using as base the SiteConfiguration model.
+    More information in https://docs.djangoproject.com/en/3.0/topics/db/models/#proxy-models
+    """
+
+    class Meta:
+        """ Set as a proxy model. """
+        proxy = True
+
+    def __unicode__(self):
+        key = getattr(settings, "USE_EOX_TENANT", "No tenant is active at the moment")
```
morenol: Shouldn't it be "EDNX_TENANT_KEY"?

eox_tenant/monkey_patch/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""
```
morenol: This is not being used

jfavellar90:
@andrey-canon I approved by mistake

eox_tenant/settings/common.py
```diff
@@ -52,6 +48,7 @@ def plugin_settings(settings):
     settings.CHANGE_DOMAIN_DEFAULT_SITE_NAME = "stage.edunext.co"
     settings.EOX_TENANT_LOAD_PERMISSIONS = True
     settings.EOX_TENANT_APPEND_LMS_MIDDLEWARE_CLASSES = False
+    settings.USE_EOX_TENANT = True
```
morenol: Why do we need this change? 

This middleware https://github.com/eduNEXT/eox-tenant/blob/be8fe48a08ba1cfd836347e9daf78e6b7d3e3d2b/eox_tenant/middleware.py#L106 is now broken because TenantSiteConfigProxy.enabled  always return True.

Probably this change is OK, but then we would need to change that middleware

andrey-canon: This new setting is not the problem, this is only for active or deactivate tenants, replace the setting USE_MICROSITES, the problem is that the middleware was designed to work when site_configuration is not enabled, and this pr is designed  to start to work with site_configuration, I'll check how to fix it

morenol:
I think that we can monkeypatch the CourseEnrollment model manager and remove the usage of this file eox_tenant/tenant_aware_functions/enrollments.py and revert this PR.

eox_tenant/middleware.py
felipemontoya: We moved this to the ready method of the apps config

eox_tenant/apps.py
```diff
@@ -51,3 +51,6 @@ def ready(self):
         """
         from eox_tenant.permissions import load_permissions
         load_permissions()
+
+        from eox_tenant.monkey_patch import load_monkey_patchs_overrides
+        load_monkey_patchs_overrides()
```
felipemontoya: Either `patches` or `patch`. I think monkey_patch_overrides is the correct version

eox_tenant/middleware.py
```diff
@@ -126,3 +93,17 @@ def process_request(self, request):
                     {'domain': domain, }
                 )
             )
+
+
+class EoxTenantCurrentSiteMiddleware(MiddlewareMixin):
+    """
+    Middleware class that define the site and its configuration.
```
felipemontoya: Perhaps we should say here that this is meant to replace django.contrib....

eox_tenant/middleware.py
```diff
@@ -126,3 +93,17 @@ def process_request(self, request):
                     {'domain': domain, }
                 )
             )
+
+
+class EoxTenantCurrentSiteMiddleware(MiddlewareMixin):
```
felipemontoya: What is you opinion on leaving the name CurrentSiteMiddleware since this is supposed to be a drop in replacement.
In common.py it will still be super clear, because of the path not being `django.contrib...` but `eox_tenant....`

eox_tenant/monkey_patch/monkey_patch_proxys.py
```diff
@@ -0,0 +1,160 @@
+"""
+Monkey patch proxys that allows to override the platform models.
```
felipemontoya: the plural of proxy is proxies. We might have to replace this all over

eox_tenant/monkey_patch/monkey_patch_proxys.py
```diff
@@ -0,0 +1,160 @@
...
+    class Meta:
+        """ Set as a proxy model. """
+        proxy = True
+
+    def __unicode__(self):
+        key = getattr(settings, "EDNX_TENANT_KEY", "No tenant is active at the moment")
+        return u"<Tenant proxy as site_configuration: {}>".format(key)
+
+    @property
+    def enabled(self):
```
felipemontoya: This method is most certainly not doing what it should be doing.

USE_EOX_TENANT is a global flag to turn the complete plugin on and off.

The real logic for this is to check for the existance of the EDNX_TENANT_KEY in the settings object

```
    def enabled(self):
        if getattr(settings, 'EDNX_TENANT_KEY', None):
            return True
        return False
```

eox_tenant/monkey_patch/monkey_patch_proxys.py
```diff
@@ -0,0 +1,160 @@
...
+        if not cls.has_configuration_set():
+            return default
+
+        cache_key = "org-value-{}-{}".format(org, val_name)
+        cached_value = cache.get(cache_key)
+
+        if cached_value:
+            return cached_value
+
+        tenant_config = TenantConfig.objects.values_list("lms_configs")
```
felipemontoya: What is the final word on the caching of this queries? How many queries do we run for the Studio /home/ page when there are many courses of different orgs?

eox_tenant/receivers_helpers.py
```diff
@@ -0,0 +1,49 @@
...
+    Reach for the configuration for a given domain.
+
+    **Arguments**
+        domain: String parameter.
+
+    **Returns**
+        configurations: dict
+        external_key: String
+
+        **Example**
```
felipemontoya: Is this config example really necessary, I think by now somebody reading the code know what to expect in the configuration dict? maybe I'm too used to this, but still

eox_tenant/receivers_helpers.py
```diff
@@ -0,0 +1,49 @@
...
+                    "BLOG":"",
+                }
+                ...
+            }
+
+            external_key = "this_is_my_key"
+    """
+    configurations, external_key = TenantConfig.get_configs_for_domain(domain)
+
+    if not (configurations and external_key):
```
felipemontoya: I think this was copied from my POC, but it is an ugly looking logic, dont you think? maybe there is a way to make it more elegant

eox_tenant/settings/aws.py
```diff
@@ -61,6 +59,9 @@ def plugin_settings(settings):  # pylint: disable=function-redefined
         settings.SITE_ID
     )
 
+    if DJANGO_CURRENT_SITE_MIDDLEWARE in settings.MIDDLEWARE_CLASSES:
```
felipemontoya: Do you think we should be doing this in common.py instead of aws.py?

eox_tenant/settings/common.py
felipemontoya: I don't think that this should be set as a default. IIRC this key USE_EOX_TENANT must be set via ansible to turn the thing on

eox_tenant/templatetags/ednx.py
```diff
@@ -7,12 +7,10 @@
...
-from eox_tenant.edxapp_wrapper.get_microsite_configuration import get_microsite
 from eox_tenant.edxapp_wrapper.branding_api import get_branding_api
-from eox_tenant.edxapp_wrapper.configuration_helpers import get_configuration_helpers
+from eox_tenant.edxapp_wrapper.site_configuration_module import get_configuration_helpers
 from eox_tenant.edxapp_wrapper.theming_helpers import get_theming_helpers
 
-microsite = get_microsite()  # pylint: disable=invalid-name
 configuration_helpers = get_configuration_helpers()
 branding_api = get_branding_api()
 theming_helpers = get_theming_helpers()
```
felipemontoya: There are a few tags below that still say microsite. Can we place deprecation notices on them?

eox_tenant/monkey_patch/monkey_patch_proxys.py
felipemontoya: Also I suggest we change the name of this file from moneky_patch/monkey_patch_proxys.py to proxies/models.py

felipemontoya:
> I think that we can monkeypatch the CourseEnrollment model manager and remove the usage of this file eox_tenant/tenant_aware_functions/enrollments.py and revert this https://github.com/eduNEXT/edunext-platform/pull/323/files.

I'm looking forward to that as well. But it will be in a different PR. Once the proxy pattern has been established and fully tested there are many proxies I'd like to implement.

eox_tenant/monkey_patch/monkey_patch_proxys.py
morenol: I think that with this change, the AvailableScreen Middleware should work as expected

eox_tenant/receivers_helpers.py
andrey-canon: you POC is a copy from the current code, but check the new way

eox_tenant/settings/aws.py
andrey-canon: to keep the current pattern IMO we should do this in aws, if we want to move this to common we should move all the middlewares and this is out of the scope of this pr

eox_tenant/templatetags/ednx.py
andrey-canon: done

eox_tenant/monkey_patch/monkey_patch_proxys.py
andrey-canon: @felipemontoya  and if we need to override another thing that is not a proxy ??

eox_tenant/middleware.py
```diff
@@ -126,3 +93,27 @@ def process_request(self, request):
...
+
+    def process_request(self, request):
+        """
+        Get the current site for a given request a set the configuration.
+        """
+        site = get_current_site(request)
+        site.configuration = TenantSiteConfigProxy()
+        request.site = site
+
+        for i in range(0, 1500):
```
morenol: Don't forget to remove this piece of code

eox_tenant/test/test_middleware.py
```diff
@@ -164,3 +120,34 @@ def test_request_triggers_not_found(
         theming_helper_mock.is_request_in_themed_site.assert_called_once()
         http_resp_not_found_mock.assert_called_once()
         self.assertEqual(result, 'redirect_response')
+
+
+class EoxTenantCurrentSiteMiddlewareTest(TestCase):
+    """
+    Test EoxTenantCurrentSiteMiddleware.
```
morenol: Test eox-tenant CurrentSiteMiddleware

jfavellar90:
@andrey-canon @felipemontoya a couple of general questions:

- Once the eox_tenant is installed, it would disable the SiteConfiguration model right? I mean, the get_value method in the proxy model only relies on the values from django.conf.settings. Even if the USE_EOX_TENANT setting is set to False, there's no fallback to use SiteConfiguration because the model is always patched.
- About the setting EDNX_USE_SIGNAL, if it's set to False, would we break the multitenancy at all? setting USE_EOX_TENANT to True, but EDNX_USE_SIGNAL to False will make that only global settings are used. I don't know if EDNX_USE_SIGNAL is going to disappear or it's going to be used rather as a debugging setting as we expect to have that value set to True

andrey-canon:
> @andrey-canon @felipemontoya a couple of general questions:
> 
> * Once the eox_tenant is installed, it would disable the SiteConfiguration model right? I mean, the get_value method in the proxy model only relies on the values from django.conf.settings. Even if the USE_EOX_TENANT setting is set to False, there's no fallback to use SiteConfiguration because the model is always patched.
> * About the setting EDNX_USE_SIGNAL, if it's set to False, would we break the multitenancy at all? setting USE_EOX_TENANT to True, but EDNX_USE_SIGNAL to False will make that only global settings are used. I don't know if EDNX_USE_SIGNAL is going to disappear or it's going to be used rather as a debugging setting as we expect to have that value set to True


@jfavellar90 
- Yes, if you install the plugin in that version SiteConfiguration will be overridden without asking 
-  EDNX_USE_SIGNAL is a setting per tenant, so if it is False,(as you say) the tenant will use the global settings, but the multitenancy behavior will not be modified at all. right now USE_EOX_TENANT activate o deactivate the multitenancy behavior.

eox_tenant/settings/aws.py
morenol: I think that this have to be enabled or disabled with a setting. Maybe it could be the setting _USE_EOX_TENANT_

eox_tenant/templatetags/ednx.py
felipemontoya: is DeprecationWarning a python global? or should it be defined somewhere?

andrey-canon: it is global, tested and it's working

PR #63, opened 2019-12-16
Add support for studio.

andrey-canon:
## Description.
Connect studio signals and remove studio conditional.

### DIFF
--- a/eox_tenant/apps.py
+++ b/eox_tenant/apps.py
@@ -49,7 +49,24 @@ class EdunextOpenedxExtensionsTenantConfig(AppConfig):
                         'signal_path': 'celery.signals.task_prerun',
                     },
                 ],
-            }
+            },
+            'cms.djangoapp': {
+                'relative_path': 'signals',
+                'receivers': [
+                    {
+                        'receiver_func_name': 'start_tenant',
+                        'signal_path': 'django.core.signals.request_started',
+                    },
+                    {
+                        'receiver_func_name': 'tenant_context_addition',
+                        'signal_path': 'celery.signals.before_task_publish',
+                    },
+                    {
+                        'receiver_func_name': 'start_async_tenant',
+                        'signal_path': 'celery.signals.task_prerun',
+                    },
+                ],
+            },
         },
     }
 
--- a/eox_tenant/test/test_signals.py
+++ b/eox_tenant/test/test_signals.py
@@ -26,16 +26,6 @@ class StartTenantSignalTest(TestCase):
     Testing the signal sent at the start of every request
     """
 
-    def test_nothing_changes_studio(self):
-        """
-        The cms should return immediately, just one call must be made it,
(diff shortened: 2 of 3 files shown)


eox_tenant/apps.py
```diff
@@ -49,7 +49,24 @@ class EdunextOpenedxExtensionsTenantConfig(AppConfig):
...
+                    {
+                        'receiver_func_name': 'finish_tenant',
+                        'signal_path': 'django.core.signals.request_finished',
+                    },
+                    {
+                        'receiver_func_name': 'clear_tenant',
+                        'signal_path': 'django.core.signals.got_request_exception',
+                    },
+                ],
+            },
```
felipemontoya: What about the async signals?

eox_tenant/test/test_signals.py
```diff
@@ -26,16 +26,6 @@ class StartTenantSignalTest(TestCase):
     Testing the signal sent at the start of every request
     """
 
-    def test_nothing_changes_studio(self):
```
felipemontoya: Can we add a test that makes sure that actually things in studio do change and thus perform as adverticed?

eox_tenant/apps.py
```diff
@@ -49,7 +49,24 @@ class EdunextOpenedxExtensionsTenantConfig(AppConfig):
...
+            'cms.djangoapp': {
+                'relative_path': 'signals',
+                'receivers': [
+                    {
+                        'receiver_func_name': 'start_tenant',
+                        'signal_path': 'django.core.signals.request_started',
+                    },
+                    {
+                        'receiver_func_name': 'finish_tenant',
```
felipemontoya: we never use this signal in the lms. Should we add it here?

eox_tenant/apps.py
```diff
@@ -49,7 +49,24 @@ class EdunextOpenedxExtensionsTenantConfig(AppConfig):
...
+                        'signal_path': 'django.core.signals.request_finished',
+                    },
+                    {
+                        'receiver_func_name': 'clear_tenant',
```
felipemontoya: we never use this signal in the lms. Should we add it here?

eox_tenant/apps.py
andrey-canon: I don't remember the reason to add this in lms, but if we don't want to add that here, because we don't use it, we should consider to clean all the unused code.

eox_tenant/apps.py
andrey-canon: same comment as above

eox_tenant/apps.py
felipemontoya: I agree. At least lets clean up this 2 signals from this and also the lms

eox_tenant/test/test_signals.py
felipemontoya: This was discussed in a different channel and discarded

PR #64, opened 2019-12-20
Fixing platform tests.

andrey-canon:
## Description
- Verify dependency fake_model
- turn off by default.

### DIFF
--- a/eox_tenant/edxapp_wrapper/backends/certificates_module_i_v1.py
+++ b/eox_tenant/edxapp_wrapper/backends/certificates_module_i_v1.py
@@ -15,4 +15,9 @@ def get_certificates_models():
 
     Return <lms.djangoapps.certificates.models>.
     """
+
+    # The following logic is necessary, because we don't want to use the certificate models as backend,
+    # when we are running the platform test, in order to avoid django migration errors.
+    # USE_EOX_TENANT is set to False by default, which allows to run the platform tests with its normal behavior,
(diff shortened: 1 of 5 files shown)


eox_tenant/edxapp_wrapper/backends/certificates_module_i_v1.py
```diff
@@ -2,13 +2,17 @@
...
+
 
 def get_certificates_models():
     """
     Backend function.
 
     Return <lms.djangoapps.certificates.models>.
     """
-    return models
+    return models if getattr(settings, 'USE_EOX_TENANT', False) else TestCertificatesModels()
```
felipemontoya: Is this line present only for the tests on the platform?

I find the logic here difficult to follow.
In the cases where we are testing eox-tenant only, we should have USE_EOX_TENANT set to true, but we will not use this backend. Is that correct?

Could you at least  make a comment explaining the necessity for this?

eox_tenant/signals.py
```diff
@@ -223,6 +223,9 @@ def start_async_tenant(sender, *args, **kwargs):  # pylint: disable=unused-argum
     See:
        https://celery.readthedocs.io/en/latest/userguide/signals.html#task-prerun
     """
+    if not getattr(base_settings, "USE_EOX_TENANT", False):
```
felipemontoya: @morenol note here that we also need to turn this on for Async servers when going to prod.

PR #66, opened 2019-12-23
Improve performance certificates.

andrey-canon:
## Description 
Improve certificates get_queryset method performance, using Q objects.

Another option is the following code but this is three times slower
```
course_ids = CourseOverview.get_all_courses(orgs=org_filter).values_list("id", flat=True)
return generated_certificates.filter(course_id__in=course_ids)
```

### DIFF
--- a/eox_tenant/tenant_wise/proxies.py
+++ b/eox_tenant/tenant_wise/proxies.py
@@ -12,6 +12,7 @@ from django.db import models
 
 from eox_tenant.edxapp_wrapper.certificates_module import get_certificates_models
 from eox_tenant.edxapp_wrapper.site_configuration_module import get_site_configuration_models
+from eox_tenant.organizations import get_organizations
 from eox_tenant.models import Microsite, TenantConfig
 
 SiteConfigurationModels = get_site_configuration_models()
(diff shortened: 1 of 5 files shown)


eox_tenant/tenant_wise/proxies.py
```diff
@@ -176,12 +176,20 @@ def get_queryset(self):
         Call parent method and filter the certificates by org.
         """
         generated_certificates = super(TenantCertificateManager, self).get_queryset()
-        excluded_courses = [
-            certificate.course_id
-            for certificate in generated_certificates
-            if certificate.course_id.org not in getattr(settings, "course_org_filter", [])
-        ]
-        return generated_certificates.exclude(course_id__in=excluded_courses)
+        org_filter = getattr(settings, "course_org_filter", [])
```
felipemontoya: Should we store this code that gets the `orgs` from the current site in a model or a reusable module?
I believe we are going to use it quite often, but I'm not sure where would the best place be.

eox_tenant/tenant_wise/proxies.py
```diff
@@ -176,12 +176,20 @@ def get_queryset(self):
...
+
+        if isinstance(org_filter, six.string_types):
+            org_filter = set([org_filter])
+
+        if org_filter:
+            q_objects = models.Q()  # Create an empty Q object to start with.
```
felipemontoya: In my opinion the line is clear enough without the comment, but it does not hurt either.

andrey-canon:
> I like the changes. I suppose this is way more efficient.
> Do we have an idea of how much faster it is?
> Have you tested this in Stage?

in stage the response time was approx 340 ms now it is approx 420 ms

eox_tenant/general_utils.py
```diff
@@ -0,0 +1,18 @@
+"""
```
felipemontoya: I would say everything is a util. Why not call this file organizations.py ?

PR #67, opened 2019-12-23
Improve performance get_value_for_org

andrey-canon:
## Description
set cache in the first call in order to make the loop once

### DIFF
--- a/eox_tenant/tenant_wise/proxies.py
+++ b/eox_tenant/tenant_wise/proxies.py
@@ -143,7 +143,7 @@ class TenantSiteConfigProxy(SiteConfigurationModels.SiteConfiguration):
                     org_filter = set([org_filter])
 
                 for organization in org_filter:
-                    if org == organization:
+                    if org == organization and not result:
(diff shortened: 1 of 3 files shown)


eox_tenant/tenant_wise/proxies.py
```diff
@@ -126,25 +126,35 @@ def get_value_for_org(cls, org, val_name, default=None):
         if cached_value:
             return cached_value
 
-        tenant_config = TenantConfig.objects.values_list("lms_configs")
-        microsite_config = Microsite.objects.values_list("values")  # pylint: disable=no-member
+        result = None
+        source = "lms_configs" if getattr(settings, "SERVICE_VARIANT", "lms") == "lms" else "studio_configs"
```
felipemontoya: We can leave this part out and we will come back to it in a PR only to make this access calls transparent to the internal location of the configs

andrey-canon:
This was closed since the pr https://github.com/eduNEXT/eox-tenant/pull/68 has been selected as  a better choice.

PR #68, opened 2019-12-26
get value for org alternative

andrey-canon:
## Description 
This is an alternative solution for https://github.com/eduNEXT/eox-tenant/pull/67 use from `django_mysql.models import JSONField` instead of `from jsonfield.fields import JSONField`

This version has been tested in stage and works.


eox_tenant/tenant_wise/proxies.py
```diff
@@ -126,25 +124,23 @@ def get_value_for_org(cls, org, val_name, default=None):
...
-        microsite_config = Microsite.objects.values_list("values")  # pylint: disable=no-member
+        user = get_current_request().user
 
-        for config in chain(tenant_config, microsite_config):
-            try:
-                current = json.loads(config[0])
-                org_filter = current.get("course_org_filter", {})
-            except IndexError:
-                continue
+        if user.is_superuser:
```
felipemontoya: We can move this to the platform, specifically the view where we render the index

eox_tenant/tenant_wise/proxies.py
```diff
@@ -126,43 +158,98 @@ def get_value_for_org(cls, org, val_name, default=None):
...
+        if not result:
+            logger.warning("Tha value for %s and org %s has not been found", val_name, org)
+            cls.set_key_to_cache(cache_key, default)
+            result = default
+
+        return result
 
     @classmethod
-    def set_key_to_cache(cls, key, value):
+    def pre_load_values_by_org(cls, val_name):
```
morenol: When will this function be called?

felipemontoya: For now we have 2 candidates.
1) when a superuser loads the home page of studio. We will do this directly from the platform, to address the fact that superusers try to load about 6k courses and thus do 12k calls to the get value by org.

2) when a superuser calls the schedule and details of a course in studio. We have noticed that this page was also speed up when we preloaded the cache.

Since calling the funcion has a high cost, we might need to check the cache first to see if it is actually necessary to do so. But that should be done also in there.
@andrey-canon we need to be aware of this when we do the PR to start calling this.

eox_tenant/tenant_wise/proxies.py
```diff
@@ -116,7 +120,35 @@ def get_value_for_org(cls, org, val_name, default=None):
...
         """
+        try:
+            return cls.__get_value_for_org(org, val_name, default)
+        except FieldError:
+            return cls.__deprecated_get_value_for_org(org, val_name, default)
+
+    @classmethod
+    def has_configuration_set(cls):
+        """
+        We always require a configuration to function, so we can skip the query
```
felipemontoya: historically, this used to mean that we where checking if there was a configuration set for the one particular site.
For performance reasons we removed this as it was making a query to the database.
Removing the query was a performance optimization that was worth the fact that we broke the implementation of the function. We can now do the real implementation without the query hit.

Could you please look at the places where this is currently being called? maybe this is worth keeping and fixing and maybe we can just remove it. Maybe there are no more calls to this in the platform and thus is only internal and we can refactor it.

eox_tenant/tenant_wise/proxies.py
```diff
@@ -126,43 +158,98 @@ def get_value_for_org(cls, org, val_name, default=None):
...
+
+        return result
+
+    @classmethod
+    def __deprecated_get_value_for_org(cls, org, val_name, default=None):
+        """
+        This method iterates over all the TenantConfig and Microsite objects, set the cache for the given
+        val_name and all organizations and returns the value for the given org.
+
+        DEPRECATED: use __get_value_for_org method instead.
```
felipemontoya: could you please leave a comment here stating why the deprecated method exists at all still. In my understanding we will never remove it since the new method requires mysql which we not always have. E.g the testing environment.

eox_tenant/tenant_wise/proxies.py
```diff
@@ -126,43 +158,98 @@ def get_value_for_org(cls, org, val_name, default=None):
...
-        Stores a key value pair in a cache scoped to the thread
+        Save in cache the values for all the organizations in TenantConfig and Microsite models.
         """
-        cache.set(
-            key,
-            value,
-            EOX_TENANT_CACHE_KEY_TIMEOUT
-        )
+        tenant_config = TenantConfig.objects.filter(
+            lms_configs__has_key=u"course_org_filter",
```
felipemontoya: We are using the filter that requires mysql 5.7+, but we have no deprecated method for this. I would suggest that we try catch this queries If there is no mysql, then the preloading will not work, but at least is not going to break the execution of the request.

eox_tenant/tenant_wise/proxies.py
morenol: Does it make sense to add this function to this PR? Right now I think that it is not being used

requirements.txt
```diff
@@ -13,6 +13,7 @@ django-fake-model==0.1.4
 django==1.11.23
 edx-opaque-keys==0.4.4
 enum34==1.1.6             # via astroid
+django-mysql==2.4.1
```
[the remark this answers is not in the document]
andrey-canon: https://github.com/eduNEXT/edunext-platform/blob/master/requirements/edx/base.in#L51

eox_tenant/tenant_wise/proxies.py
felipemontoya: Yes of course. As soon as this is merged we will be merging as well the PR that uses it on edx-platform. Idk if the PR exists already, but we are already working on it.

eox_tenant/tenant_wise/proxies.py
andrey-canon: @felipemontoya  I have checked it and we can remove it, we only use it in the methods  get_all_orgs, __get_value_for_org and __deprecated_get_value_for_org, and those will be called if the CurrentSiteMiddleware was replace https://github.com/eduNEXT/eox-tenant/blob/cb28cce01203d4c65e46471863e5a89d31a7cccb/eox_tenant/settings/aws.py#L65  and there we have the same condition

PR #74, opened 2020-01-14
Add the capacity to read the values at the proxy

felipemontoya:
This PR adds a method so that the SiteConfigProxy also allows the platform code to access the `.values` property of the SiteConfig.

This is only used at helpers.has_configuration_override but impacts many areas of the code. Until the removal of microsites it was not a problem due to how the microsites backend handled the defaults, but now it is more standard


felipemontoya:
@luismorenolopera @morenol I'm going to merge this with only one approval, but please keep this change in mind fyi

PR #77, opened 2020-01-24
Monkey patching by djangoapp.

andrey-canon:
## Description
This allows to use any proxy in a whole djangoapp.

### DIFF
--- a/eox_tenant/tenant_wise/__init__.py
+++ b/eox_tenant/tenant_wise/__init__.py
@@ -3,6 +3,7 @@ Eox Tenant Wise.
 ==================
 This module makes it possible to override the some platform Models using new proxy models.
 """
+import inspect
 from importlib import import_module
 
 import six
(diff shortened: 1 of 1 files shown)


eox_tenant/tenant_wise/__init__.py
```diff
@@ -26,25 +27,29 @@ def load_tenant_wise_overrides():
...
 
 
-def set_as_proxy(modules, model, proxy):
+def set_as_proxy(model, proxy, djangoapp='', modules=None):
     """
     Helper to patch a loaded module with a proxy object that has all the Tenant wise properties.
     """
-    if isinstance(modules, six.string_types):
-        modules = set([modules])
+    if djangoapp:
```
felipemontoya: Could it happen that there is no djangoapp? Why is the default '' ?

andrey-canon: the default is an empty string in order to indicate that the djangoapp value should be a string, and yes, there is no djangoapp when the method only is used to patch specific modules

felipemontoya: I cant exactly put my finger on it, but this functions looks weird.
Part of it is the empty string as default and the other part is that is has 2 different implementations behind the `if` statements.

Also, the djangoapp parameter does not have to be a djangoapp, does it? It is only a path to a package. Only that this time we expect this 'package' to have subpackages.

PR #78, opened 2020-01-24
Allow to make regressions over a model.

andrey-canon:
## Description 
Modify  the __class__ field in order to use the previous model.

### DIFF
--- a/eox_tenant/tenant_wise/context_managers.py
+++ b/eox_tenant/tenant_wise/context_managers.py
@@ -9,16 +9,18 @@ logger = logging.getLogger(__name__)
 
 
 @contextmanager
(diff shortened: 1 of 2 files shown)


mariajgrimaldi:
Hey! What happened with this PR? @felipemontoya

felipemontoya:
I actually have no idea.
@andrey-canon is this something we still want to merge? we are not using this in prod, but we have [this commit](https://github.com/eduNEXT/edunext-platform/commit/e34418c97af796c2655454e67eddfb19201d3b96) currently in the code.
Does this solve it?
Do we continue or should we close it? If we decide to continue we must test this quite well in staging

PR #79, opened 2020-03-03
Search in JSONFields for listview in admin

luismorenolopera:
- Add a search in the admin for fields that are JSONField, for TenantConfig and Microsite models.
- Fix a 500 error that was generated when trying to search these views.

### DIFF
--- a/eox_tenant/models.py
+++ b/eox_tenant/models.py
@@ -31,6 +31,55 @@ else:
     from jsonfield.fields import JSONField
 
 
+class JsonSearchMixin(object):
+    """
+    Mixin to search for values in JSONField.
+    """
+    def filter_on_json_fields(self, search_term, fields):
+        """
+        Look for the value in the given field list, the fields must be of type JSONField.
+
+        Args:
+            search_term: String.
+            fields: List of Strings.
+        Returns:
+            Queryset list of the invoking class.
+        """
+        if search_term == '':
+            return self.model.objects.none()
+
+        sql = """SELECT id FROM {table} WHERE """ \
+              .format(table=self.model._meta.db_table)  # pylint: disable=protected-access
+        sql = self.add_fields(sql, fields)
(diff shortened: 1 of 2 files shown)


eox_tenant/models.py
```diff
@@ -108,6 +108,29 @@ def get_value_for_org(cls, org, val_name):
...
+            return cls.objects.none()  # pylint: disable=no-member
+
+        sql = """
+            SELECT id FROM {table}
+            WHERE JSON_EXTRACT(`values`,"$.*") LIKE %s""" \
+            .format(table=cls._meta.db_table)  # pylint: disable=no-member
+        with connection.cursor() as cursor:
+            cursor.execute(sql, ['%' + search_term + '%'])
+            rows = cursor.fetchall()
+        rows = [row[0] for row in rows]
```
luismorenolopera: @felipemontoya  Q: Can you do something like https://stackoverflow.com/questions/37205793/django-values-list-vs-values to have a flat list from the query?

R: I didn't find something similar for a RAW query

eox_tenant/models.py
```diff
@@ -108,6 +108,29 @@ def get_value_for_org(cls, org, val_name):
 
         return None
 
+    @classmethod
```
jfavellar90: @luismorenolopera I have a suggestion related to code organization. I notice the code below is similar to the one you added on the **TenantConfig** model, what changes is the model fields to use in the query. In this order of ideas, we can:

- Move this method to a model manager (named **filter_on_json_fields** or something much better)
- This method should be able to create the query for a variable number of fields
- The fields to use in the query can be defined via a model attribute, a model method, a manager method or accept a list of fields as a param on the **filter_on_json_fields** method
- Make involved models use the manager

This could be even a mixin to extend a manager. Does it make sense?

jfavellar90:
LGTM :+1:

PR #80, opened 2020-03-16
Added Basic view to show the working version and the exact git commit

luismorenolopera:


eox_tenant/views.py
```diff
@@ -0,0 +1,37 @@
...
+            ["git", "rev-parse", "HEAD"], cwd=working_dir))
+    except CalledProcessError:
+        git_data = ""
+
+    response_data = {
+        "version": eox_tenant.__version__,
+        "name": "eox-tenant",
+        "git": git_data.rstrip('\r\n'),
+    }
+    return HttpResponse(
```
andrey-canon: we can use JsonResponse

eox_tenant/views.py
```diff
@@ -0,0 +1,37 @@
...
+
+from __future__ import unicode_literals
+
+import json
+from os.path import dirname, realpath
+from subprocess import check_output, CalledProcessError
+
+from django.http import HttpResponse
+from django.utils import six
+
```
andrey-canon: you can remove this blank space

eox_tenant/apps.py
```diff
@@ -12,6 +12,13 @@ class EdunextOpenedxExtensionsTenantConfig(AppConfig):
     verbose_name = "Edunext Openedx Multitenancy."
 
     plugin_app = {
+        'url_config': {
+            'lms.djangoapp': {
+                'namespace': 'eox-tenant',
+                'regex': r'^eox-tenant/',
+                'relative_path': 'urls',
+            }
```
andrey-canon: Add trailing comma

eox_tenant/views.py
```diff
@@ -0,0 +1,37 @@
...
+
+import eox_tenant
+
+
+def info_view(request):  # pylint: disable=unused-argument
```
andrey-canon: @felipemontoya  this is a view that anyone can see, do we want that behavior? or maybe we should restrict this information

eox_tenant/views.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    Basic view to show the working version and the exact git commit of the
+    installed app.
+    """
+    try:
+        working_dir = dirname(realpath(__file__))
+        git_data = six.text_type(check_output(
```
andrey-canon: imo the line doesn't need this break, it's not long enough

andrey-canon:
@luismorenolopera  could you add tests ?

eox_tenant/views.py
luismorenolopera: Without the line break it exceeds 80 lines

andrey-canon: we use 120 as the max number of characters https://github.com/eduNEXT/eox-tenant/blob/master/setup.cfg#L17

eox_tenant/test/test_views.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+
+    def test_view_info_accesible(self):
+        """
+        Should get a successful answer
+        """
+        response = self.client.get('/eox-info')
+        self.assertEqual(response.status_code, 200)
+
+    def test_view_info_response_data(self):
```
andrey-canon: Could you consider the exception case ?

eox_tenant/views.py
felipemontoya: Currently for the eox-core view it is just like this. Anyone can view it.

There is an inherent risk if one of our versions has a bug or known vulnerability and it is not yet patched for the instance. But we accepted this for eox-core I don't see why we can't accept it for tenant.
Now, If you know about a recommended practice not to do it this way, please speak up and we can change this behaviour

andrey-canon:
@luismorenolopera why this was merged without any approvals and with failing tests ???

felipemontoya:
@andrey-canon I did not noticed before there where failing tests and outstanding issues not addressed. We should revert this PR and @luismorenolopera should go through the final stage of the PR again.

felipemontoya:
@mariajgrimaldi can you take over this PR please?

PR #81, opened 2020-04-14
Add basic viewset endpoints

luismorenolopera:
This adds a simple crud API, only super users have permissions.


eox_tenant/api/v1/viewsets.py
```diff
@@ -0,0 +1,34 @@
...
+
+from rest_framework import viewsets
+from rest_framework.permissions import IsAdminUser
+
+
+class MicrositeViewSet(viewsets.ModelViewSet):
+    """
+    Microsite viewset.
+    """
+    permission_classes = [IsAdminUser]
```
felipemontoya, line 14: Can we make this Super User only?

eox_tenant/api/v1/viewsets.py
```diff
@@ -0,0 +1,34 @@
...
+    serializer_class = MicrositeSerializer
+    queryset = Microsite.objects.all()
+
+
+class TenantConfigViewSet(viewsets.ModelViewSet):
+    """
+    TenantConfig viewset.
+    """
+    permission_classes = [IsAdminUser]
```
felipemontoya, line 23: super user

eox_tenant/api/v1/viewsets.py
```diff
@@ -0,0 +1,34 @@
...
+    serializer_class = TenantConfigSerializer
+    queryset = TenantConfig.objects.all()
+
+
+class RouteViewSet(viewsets.ModelViewSet):
+    """
+    Route viewset.
+    """
+    permission_classes = [IsAdminUser]
```
felipemontoya, line 32: super user

felipemontoya:
I left some inline comments, and also the tests have to pass.
Probably we need to add tests so that testing coverage does not go down

eox_tenant/api/v1/routers.py
```diff
@@ -0,0 +1,10 @@
+"""
+Routers file for eox_tenant api v1.
+"""
+from rest_framework import routers
+from eox_tenant.api.v1.viewsets import MicrositeViewSet, TenantConfigViewSet, RouteViewSet
```
andrey-canon, line 5: bad sort and it's missing a space before this line

eox_tenant/api/v1/serializers.py
```diff
@@ -0,0 +1,50 @@
+"""
+Serializers file for eox_tenant api v1.
+"""
+from rest_framework import serializers
+
+from eox_tenant.models import Microsite, TenantConfig, Route
```
andrey-canon, line 6: bad sort

eox_tenant/api/v1/urls.py
```diff
@@ -0,0 +1,12 @@
+"""
+Version 1 URL Configuration
+
+The `urlpatterns` list routes URLs to views. For more information please see:
+    https://docs.djangoproject.com/en/1.11/topics/http/urls/
+"""
+from django.conf.urls import include, url
+from eox_tenant.api.v1.routers import router
```
andrey-canon, line 8: blank space before

eox_tenant/api/v1/viewsets.py
```diff
@@ -0,0 +1,34 @@
+from eox_tenant.api.v1.serializers import (MicrositeSerializer,
```
andrey-canon, line 1: bad sort, sort in this way 
Future
Python Standard Library
Third Party
Current Python Project

we don't have an specific way to this but I prefer the following way 
image

eox_tenant/urls.py
```diff
@@ -1,9 +1,9 @@
 """eox-tenant urls"""
 
-from django.conf.urls import url
+from django.conf.urls import include, url
```
andrey-canon, line 3: blank space after this

eox_tenant/api/v1/urls.py
luismorenolopera: done

eox_tenant/urls.py
luismorenolopera: done

eox_tenant/api/v1/tests/test_microsites.py
```diff
@@ -0,0 +1,130 @@
...
+
+from django.contrib.auth.models import User
+from mock import patch
+from rest_framework import status
+from rest_framework.test import APIClient, APITestCase
+
+from eox_tenant.models import Microsite
+
+
+class MicrositeAPITest(APITestCase):
```
andrey-canon, line 13: could we test the format of the response ? I know we don't have a established format yet but this could help to ensure that the response will not change in the future by accident, if we are waiting a error field in a 400 response it could be nice if we verify that.
But I'm not sure 100 per cent, please let me know your thoughts

eox_tenant/api/v1/tests/test_microsites.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    patch_permissions = patch('eox_tenant.api.v1.permissions.EoxTenantAPIPermission.has_permission', return_value=True)
+
+    def setUp(self):
+        """
+        setup.
+        """
+        super(MicrositeAPITest, self).setUp()
+        self.api_user = User(1, 'test@example.com', 'test')
+        self.url = '/api/v1/microsites/'
```
andrey-canon, line 25: could you use reverse instead ?

eox_tenant/api/v1/tests/test_microsites.py
```diff
@@ -0,0 +1,130 @@
+"""
```
andrey-canon, line 1: your docstrings 'must **be** ...'

eox_tenant/api/v1/tests/test_microsites.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @patch_permissions
+    def test_create_microsite(self, _):
+        """
+        Must be create new Microsite.
+        """
+        data = {
+            'key': 'test_fake_key',
+            'subdomain': 'subdomain.localhost',
+            'values': {'key': 'value'}
```
andrey-canon, line 58: trailing comma

eox_tenant/api/v1/tests/test_microsites.py
```diff
@@ -0,0 +1,130 @@
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    @patch_permissions
+    def test_invalid_update_microsite(self, _):
+        """
+        Must be fail when try update a Microsite.
+        """
+        # Test empty request
+        data = {}
+        response = self.client.put('{url}{id}/'.format(url=self.url, id=self.microsite_example.pk),
```
andrey-canon, line 102: same comment as above

eox_tenant/api/v1/tests/test_microsites.py
```diff
@@ -0,0 +1,130 @@
...
+                                   data=data, format='json')
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        # Test with invalid request (all fields are required)
+        data = {
+            "key": "new_key",
+            "subdomain": "subdomain.new"
```
andrey-canon, line 108: trailing comma

eox_tenant/api/v1/tests/test_microsites.py
```diff
@@ -0,0 +1,130 @@
...
+from rest_framework.test import APIClient, APITestCase
+
+from eox_tenant.models import Microsite
+
+
+class MicrositeAPITest(APITestCase):
+    """
+    Microsite API TestCase.
+    """
+    patch_permissions = patch('eox_tenant.api.v1.permissions.EoxTenantAPIPermission.has_permission', return_value=True)
```
andrey-canon, line 17: could you check if there is a way to do this in the class level, in order to avoid to use the decorator in every case and the useless parameter `_`.  I think we have to test where the user doesn't have the right permissions...

eox_tenant/api/v1/tests/test_microsites.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @patch_permissions
+    def test_valid_update_microsite(self, _):
+        """
+        Must be update Microsite.
+        """
+        data = {
+            "key": "new_key",
+            "subdomain": "subdomain.new",
+            "values": {"key": "updated"}
```
andrey-canon, line 89: trailing comma

eox_tenant/api/v1/tests/test_microsites.py
```diff
@@ -0,0 +1,130 @@
...
+                                   data=data, format='json')
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    @patch_permissions
+    def test_valid_patch_microsite(self, _):
+        """
+        Must be update using patch method a Microsite.
+        """
+        data = {"key": "new_patch_key"}
+        response = self.client.patch('{url}{id}/'.format(url=self.url, id=self.microsite_example.pk),
```
andrey-canon, line 120: same comment

eox_tenant/api/v1/tests/test_microsites.py
luismorenolopera: done

eox_tenant/api/v1/tests/test_microsites.py
luismorenolopera: we can do something like [this](https://stackoverflow.com/questions/6307761/how-to-decorate-all-functions-of-a-class-without-typing-it-over-and-over-for-eac). but if we want to create a test that does not have the permissions I think it would be a problem.

eox_tenant/api/v1/permissions.py
```diff
@@ -0,0 +1,15 @@
+"""
```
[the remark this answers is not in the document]
felipemontoya: This is a relative broad permission, I'm fine with this here or the top folder.

eox_tenant/api/v1/serializers.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        model = TenantConfig
+        fields = '__all__'
+
+
+class RouteSerializer(serializers.ModelSerializer):
+    """
+    Serializer class for Route model.
+    """
+
```
felipemontoya, line 44: Is this emtpy on purpose?

luismorenolopera: yes

andrey-canon: Insert a blank line after all docstrings (one-line or multi-line) that document a class -- generally speaking, the class's methods are separated from each other by a single blank line, and the docstring needs to be offset from the first method by a blank line. https://www.python.org/dev/peps/pep-0257/
indeed it's missing in in the other classes of this file

eox_tenant/api/v1/tests/test_microsites.py
luismorenolopera: this functionality is tested in JSONRenderer, It would be repetitive

PR #83, opened 2020-05-18
Added setup for isort in setup.cfg

mariajgrimaldi:
This PR adds settings in setup.cfg for using _isort_ in the project, this allows ordering the imports as follows:

1. Future imports
2. Python Standard Libraries
3. Django core
4. Third-party libraries (related or not to Django)
5. First party libraries (that is, our project’s imports)
6. Local imports

I used this configuration to sort all eox_tenant files, then I ran make run-quality-test and got 10/10.

Also, what do you think about the settings?

See references:
[Orginize imports reference](https://simpleisbetterthancomplex.com/packages/2016/10/08/isort.html)
[Isort settings](https://github.com/timothycrosley/isort/wiki/isort-Settings)
[Section settings](https://github.com/timothycrosley/isort#custom-sections-and-ordering)

@felipemontoya 
@morenol 
@andrey-canon

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -294,3 +294,9 @@ disable =
 	old-style-class,
 	inconsistent-return-statements
 
+[isort]
+known_first_party = eox_tenant
+include_trailing_comma = True
+indent = '    '
+line_length = 120
+multi_line_output = 3


morenol:
@mariajgrimaldi 
I think that we should add also these options:
```
include_trailing_comma = True
indent = '    '
line_length = 120
multi_line_output = 3
```

setup.cfg
```diff
@@ -294,3 +294,7 @@ disable =
 	old-style-class,
 	inconsistent-return-statements
 
+[isort]
+known_first_party = eox_tenant
+sections = FUTURE,STDLIB,DJANGO,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
+skip = backends
```
andrey-canon, line 300: why are you skipping the backendsfolder ? or this line has another purpose ??

setup.cfg
```diff
@@ -294,3 +294,7 @@ disable =
 	old-style-class,
 	inconsistent-return-statements
 
+[isort]
+known_first_party = eox_tenant
+sections = FUTURE,STDLIB,DJANGO,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
```
andrey-canon, line 299: This is the default value, this is not necessary

mariajgrimaldi:
> @mariajgrimaldi
> I think that we should add also these options:
> 
> ```
> include_trailing_comma = True
> indent = '    '
> line_length = 120
> multi_line_output = 3
> ```

andrey-canon:
@mariajgrimaldi you shouldn't skip the backends folder, the issue is the following 

1. Probably you are running something similar to `isort --recursive ./eox_tenant/` to test the configurations.
2. you will get this change in https://github.com/eduNEXT/eox-tenant/blob/master/eox_tenant/edxapp_wrapper/backends/edx_auth_i_v1.py#L2
image
3. you run `make run-quality-test` and get this 
image

the `# pylint: disable=import-error` should keep in the same line like this 
image

The problem is not the isort configuration, it is that you remove the pylint disable expression when you test the isort command

mariajgrimaldi:
I fixed the PR using your comments! Thank you! @morenol @andrey-canon

PR #95, opened 2020-06-17
Add changelog

morenol:
Add changelog body to eox-tenant in order to create a pypi relase.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,0 +1,71 @@
+Change Log
+----------
+
+..
+   All enhancements and patches to eox-tenant will be documented
+   in this file.  It adheres to the structure of http://keepachangelog.com/ ,
+   but in reStructuredText instead of Markdown (for ease of incorporation into
+   Sphinx documentation and the PyPI description).
+   
+   This project adheres to Semantic Versioning (http://semver.org/).
+.. There should always be an "Unreleased" section for changes pending release.
+Unreleased
+~~~~~~~~~~
+
+* Add tests in python 3.5
+
+[2.5.0] - 2020-01-03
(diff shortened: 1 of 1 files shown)


Squirrel18:
I think we should add the full changelog including all versions. What do you think? @morenol

morenol:
> I think we should add the full changelog including all versions. What do you think? @morenol
I think that that would be ideal but I am not sure how easy is to add them and if the effort to add them is worth it

Squirrel18:
I'm pretty sure it's not easy, but I think we should do it anyway because what is the point of adding a changelog file if we are not going to document the older releases (35 releases)? Does the incoming changelog addition have to contain all the features so far? Maybe we can add a changelog record for only major versions. @morenol

morenol:
> I'm pretty sure it's not easy, but I think we should do it anyway because what is the point of adding a changelog file if we are not going to document the older releases (35 releases)? Does the incoming changelog addition have to contain all the features so far? Maybe we can add a changelog record for only major versions. @morenol

The point is  to start report the changes for the new versions since they will be on pypi

morenol:
@Squirrel18 I tried to add to the changelog some of the changes for the previous version, I might need help to add the other ones.

CHANGELOG.rst
```diff
@@ -0,0 +1,74 @@
...
+[1.3.0] - 2019-12-12
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Added
+_____
+
+* Support multitenancy in async process.
+
+
+
```
Squirrel18, line 74: Unnecessary blank lines.

Squirrel18:
Thanks @morenol  Maybe @felipemontoya or @andrey-canon could help add more changelog versions to the file.

PR #96, opened 2020-06-17
Use circleci matrix

morenol:
Use circleci matrix to run the tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,23 +1,33 @@
-# Python CircleCI 2.0 configuration file
+# Python CircleCI 2.1 configuration file
 #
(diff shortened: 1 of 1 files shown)


luismorenolopera:
@morenol we don't need to test with python3.6 anymore?

morenol:
> @morenol we don't need to test with python3.6 anymore?

I added them again, though I think that we dont need them at all. Instead we should add the tests for python3.8, but I am not adding them for two reasons: 

* Django1.11 is not compatible with python3.8 (oficially) and we haven't done yet anything to test using different version of django.
* And the tests with python3.8 are failing

PR #99, opened 2020-07-03
Dam/production settings

Alec4r:


eox_tenant/apps.py
```diff
@@ -23,12 +23,14 @@ class EdunextOpenedxExtensionsTenantConfig(AppConfig):
...
                 'common': {'relative_path': 'settings.common'},
-                'aws': {'relative_path': 'settings.aws'},
+                'aws': {'relative_path': 'settings.production'},
+                'production': {'relative_path': 'settings.production'},
             },
             'cms.djangoapp': {
                 'test': {'relative_path': 'settings.test'},
                 'common': {'relative_path': 'settings.common'},
-                'aws': {'relative_path': 'settings.aws'},
+                'aws': {'relative_path': 'settings.production'},
```
morenol, line 32: I think that this is counterintuitive, maybe it would be better to have an aws.py file that imports all from production.py

eox_tenant/settings/production.py
```diff
@@ -67,14 +73,19 @@ def plugin_settings(settings):  # pylint: disable=function-redefined
...
 
-    if DJANGO_CURRENT_SITE_MIDDLEWARE in settings.MIDDLEWARE_CLASSES and getattr(settings, 'USE_EOX_TENANT', False):
-        settings.MIDDLEWARE_CLASSES[settings.MIDDLEWARE_CLASSES.index(DJANGO_CURRENT_SITE_MIDDLEWARE)] = 'eox_tenant.middleware.CurrentSiteMiddleware'  # pylint: disable=line-too-long
+    # Validate if MIDDLEWARE_CLASSES attribute exists in the environment
+    middleware = 'MIDDLEWARE_CLASSES'
+    if not hasattr(settings, middleware):
+        middleware = 'MIDDLEWARE'
+
+    if DJANGO_CURRENT_SITE_MIDDLEWARE in getattr(settings, middleware) and getattr(settings, 'USE_EOX_TENANT', False):
+        aux = getattr(settings, middleware)
```
andrey-canon, line 82: @Alec4r  aux, really ?

eox_tenant/apps.py
andrey-canon: @morenol @felipemontoya  what do you think if we remove the aws file, I mean no more hawthorn support ?

morenol: Yes, I think that we can do that, in any case we should add an entry in the changelog

eox_tenant/settings/production.py
```diff
@@ -67,14 +73,19 @@ def plugin_settings(settings):  # pylint: disable=function-redefined
         settings.TENANT_WISE_ALLOWED_PROXIES
     )
 
-    if DJANGO_CURRENT_SITE_MIDDLEWARE in settings.MIDDLEWARE_CLASSES and getattr(settings, 'USE_EOX_TENANT', False):
-        settings.MIDDLEWARE_CLASSES[settings.MIDDLEWARE_CLASSES.index(DJANGO_CURRENT_SITE_MIDDLEWARE)] = 'eox_tenant.middleware.CurrentSiteMiddleware'  # pylint: disable=line-too-long
+    # Validate if MIDDLEWARE_CLASSES attribute exists in the environment
+    middleware = 'MIDDLEWARE_CLASSES'
```
jfavellar90, line 77: Checking this change, my question is if we need to make the eox-tenant middlewares inherit from `django.utils.deprecation.MiddlewareMixin` (https://docs.djangoproject.com/en/2.2/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware). I checked on Juniper edx-platform some middlewares, and all of them inherit from that class

Alec4r: @jfavellar90 the "eox_tenant.middleware.CurrentSiteMiddleware" is working on juniper and the older version, the other middlewares in EOX_TENANT_MIDDLEWARES require some changes, but I think that is better to scale this to the product team.

eox_tenant/apps.py
Alec4r: @morenol so, won't we support to ironwood?

eox_tenant/settings/production.py
morenol: CurrentSiteMiddleware already has that Mixin, I also think that we need to add for the other middlewares

eox_tenant/apps.py
morenol: No, Irowood also uses the production.py setting

felipemontoya:
I am intrigued by the change from aws->production. I imagined that we would rather add "production" with all the regular changes and leave "aws" inheriting from prod.

That is what we do in tagging:
https://github.com/eduNEXT/eox-tagging/tree/master/eox_tagging/settings

Also we need to do this same thing in eox-core:
https://github.com/eduNEXT/eox-core/tree/master/eox_core/settings

morenol:
Add an entry with the description of this change to the unreleased section of the changelog

morenol:
> I am intrigued by the change from aws->production. I imagined that we would rather add "production" with all the regular changes and leave "aws" inheriting from prod.
> 

In this case, I think that we would need to be more carefull and modify the logic of those files because in Ironwood, both settings are called so it is possible that we would be duplicating the middlewares

Alec4r:
> > I am intrigued by the change from aws->production. I imagined that we would rather add "production" with all the regular changes and leave "aws" inheriting from prod.
> 
> @felipemontoya
> In this case, I think that we would need to be more carefull and modify the logic of those files because I think that in Ironwood both settings are called so it is possible that we would be duplicating the middlewares

@jfavellar90 I think that We can remove the aws file, with this, We won't have problems with duplicate middleware in ironwood.

jfavellar90:
@Alec4r or append the middlewares only if they are not present in the `MIDDLEWARE` setting?

Alec4r:
> @Alec4r or append the middlewares only if they are not present in the `MIDDLEWARE` setting?

Yes, We can do that, but... remember, We need to validate if exists MIDDLEWARE or MIDDLEWARE_CLASSES and just then We will able to validate if exist, but ... We really need to continue support hawthorn?, SAAS is in Ironwood and We will migrate to juniper in a few months... hawthorn is an old version.

morenol:
I agree with @Alec4r, unless we have a client using Hawthorn + eox-tenant, I think that there will not be problems if we just remove that setting/file.

felipemontoya:
No client is using the latest version of eox-tenant in hawthorn. I was mostly worried about ironwood but in there we are already using "production" and it is only due to https://github.com/eduNEXT/edx-platform/blob/open-release/ironwood.master/lms/envs/production.py#L1112 that it works.
I agree we should remove it.

CHANGELOG.rst
```diff
@@ -12,7 +12,7 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
-*
+* Added juniper support
```
andrey-canon, line 15: remove hawthorn support ?? @Alec4r

PR #103, opened 2020-09-22
Add tenant organizations

andrey-canon:
## Description
- Remove django_mysql jsonfields and use an alternative model called TenantOrganization
- Add synchronization command 
- Add receiver method to create organizations


mariajgrimaldi (approved):

eox_tenant/management/commands/synchronize_organizations.py
```diff
@@ -0,0 +1,64 @@
...
+            queryset = model.objects.all()
+
+            LOGGER.info("Synchronize {} {} registers.".format(
+                len(queryset),
+                valid_model,
+            ))
+
+            for instance in queryset:
+                synchronize_tenant_organizations(instance)
+                sleep(1)
```
felipemontoya, line 62: Can we make this value an override with default 0.2 seconds?

eox_tenant/models.py
```diff
@@ -25,59 +28,25 @@ def check_mysql_version():
...
+        app_label = "eox_tenant"
 
-
-class MicrositeManager(models.Manager, JsonSearchMixin):
-    """
-    Custom manager for Microsites model.
-    """
-    pass
+    def __str__(self):
+        return "<Tenant organization: {}>".format(self.name)
```
felipemontoya, line 49: For readability can we make this shorter? <Org: {}> ?

eox_tenant/utils.py
```diff
@@ -40,3 +43,29 @@ def move_signupsource(old_domain, new_domain):
         count = UserSignupSource.objects.filter(site=old_domain).update(site=new_domain)
 
         log.info("Updated {} SignupSources from {} to {}.".format(count, old_domain, new_domain))
+
+
+def synchronize_tenant_organizations(instance):
```
felipemontoya, line 48: So far this method only adds new orgs, do we have anything to remove old orgs and purge them?

eox_tenant/models.py
```diff
@@ -95,8 +64,8 @@ class Microsite(models.Model):
     """
     key = models.CharField(max_length=63, db_index=True)
     subdomain = models.CharField(max_length=127, db_index=True)
-    values = JSONField(null=False, blank=True)
-    objects = MicrositeManager()
+    values = JSONField(blank=True, default={}, load_kwargs={'object_pairs_hook': collections.OrderedDict})
```
felipemontoya, line 67: Does this change in any way imply that we need to change the data in the database?

felipemontoya (changes requested):
Im wondering some things:

- should we add a data migration for the first sync?
- should we make it fail less fatally if the migrations have not been run?

Also:
- could you please add a note in the readme about the synchronize_organizations command?
- about the admin page for the orgs:
  - could you add a text filter? both on the name of the org, and the routes/microsites domains or keys
  - can you make the list of the orgs display the tenants/microsites where it is linked?
  - on the admin page, make it only display the string of the org, not the __repr__

PR #104, opened 2020-09-25
[JUNIPER LD-9]Override SiteConfiguration in contenstore.utils

andrey-canon:
## Description
This is part of LD-9, this change will avoid to include configuration_helpers in the contentstore.utils file


felipemontoya (approved):

felipemontoya:
I approved it, but I'm not sure why are we skipping the contentstore.utils when it is the LMS

mariajgrimaldi (approved):

PR #110, opened 2020-11-19
Fix proxy compatibility issue in juniper

mariajgrimaldi:
This PR adds site_values as property to SiteConfiguration proxy, this is done due to this change: https://github.com/edx/edx-platform/pull/23306/files

Thank you for the help! @andrey-canon

### DIFF
--- a/eox_tenant/tenant_wise/proxies.py
+++ b/eox_tenant/tenant_wise/proxies.py
@@ -89,6 +89,19 @@ class TenantSiteConfigProxy(SiteConfigurationModels.SiteConfiguration):
         """
         pass
 
+    @property
+    def site_values(self):
+        """
+        Returns the raw values of the loaded settings. This version works with juniper releases.
+        """
+        return self.values
+
+    @site_values.setter
+    def site_values(self, value):
+        """
+        We ignore the setter since this is a read proxy. This version works with juniper releases.
+        """
+
     def save(self, *args, **kwargs):
         """
         Don't allow to save TenantSiteConfigProxy model in database.


eox_tenant/tenant_wise/proxies.py
```diff
@@ -89,6 +89,22 @@ def values(self, value):
...
+        if self.enabled:
+            return vars(settings._wrapped)  # pylint: disable=protected-access
+        return {}
+
+    @site_values.setter
+    def site_values(self, value):
+        """
+        We ignore the setter since this is a read proxy. This version works with juniper releases.
+        """
+        pass
```
andrey-canon, line 106: it is not necessary

eox_tenant/tenant_wise/proxies.py
```diff
@@ -89,6 +89,22 @@ def values(self, value):
         """
         pass
 
+    @property
+    def site_values(self):
+        """
+        Returns the raw values of the loaded settings. This version works with juniper releases.
+        """
+        if self.enabled:
```
andrey-canon, line 97: you can return values

andrey-canon (changes requested):

eox_tenant/tenant_wise/proxies.py
mariajgrimaldi: I didn't know, thanks!

eox_tenant/tenant_wise/proxies.py
mariajgrimaldi: ... right

andrey-canon (approved):

PR #115, opened 2021-01-18
Add method validate_bearer_token to EoxTenantOAuth2Validator

vad33r:
This PR adds the method validate_bearer_token and modifies it to validate the actual url and the result from redirect_uri_allowed.

### DIFF
--- a/eox_tenant/validators.py
+++ b/eox_tenant/validators.py
@@ -25,6 +25,6 @@ class EoxTenantOAuth2Validator(EdxOAuth2Validator):
         """Validates the token if the current url is in redirect_uris list."""
 
         if request.auth.application.redirect_uri_allowed(request.build_absolute_uri('/')):
-            super(EoxTenantOAuth2Validator, self).validate_bearer_token(token, scopes, request)
-            return True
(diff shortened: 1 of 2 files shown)


eox_tenant/validators.py
```diff
@@ -16,3 +16,11 @@ def save_bearer_token(self, token, request, *args, **kwargs):
...
         if request.auth.application.redirect_uri_allowed(request.build_absolute_uri('/')):
             super(EoxTenantOAuth2Validator, self).save_bearer_token(token, request, *args, **kwargs)
+
+    def validate_bearer_token(self, token, scopes, request):
+        """Validates the token if the current url is in redirect_uris list."""
+
+        if request.auth.application.redirect_uri_allowed(request.build_absolute_uri('/')):
+            super(EoxTenantOAuth2Validator, self).validate_bearer_token(token, scopes, request)
+            return True
+        return False
```
andrey-canon, line 26:
```suggestion
        if request.auth.application.redirect_uri_allowed(request.build_absolute_uri('/')):
            return super(EoxTenantOAuth2Validator, self).validate_bearer_token(token, scopes, request)

        return False
```

PR #123, opened 2021-05-05
feat!: Drop python 2.7 support

andrey-canon:
## Description
Drop python 2.7 support in favor of python 3.8

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,11 +14,14 @@ workflows:
               only: /v?[0-9]+(\.[0-9]+)*/
           matrix:
             parameters:
-              python_version: ["2.7", "3.5"]
-              django_version: ["django111", "django22"]
+              python_version: ["3.5", "3.8"]
+              debian_version: ["stretch", "buster"]
             exclude:
-              - python_version: "2.7"
-                django_version: "django22"
+              - python_version: "3.5"
+                debian_version: "buster"
+              - python_version: "3.8"
+                debian_version: "stretch"
+
       - pypi:
           requires:
             - test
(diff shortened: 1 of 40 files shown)


.circleci/config.yml
```diff
@@ -14,11 +14,14 @@ workflows:
               only: /v?[0-9]+(\.[0-9]+)*/
           matrix:
             parameters:
-              python_version: ["2.7", "3.5"]
-              django_version: ["django111", "django22"]
+              python_version: ["3.5", "3.8"]
+              debian_version: ["stretch", "buster"]
```
mariajgrimaldi, line 18: why this?

mariajgrimaldi (approved):

.circleci/config.yml
andrey-canon: that is the debian version, previous these changes the version was stretch by default and that version only supports until python 3.5  hence it's required a new debian version which supports python 3.8, besides there is no a image called 3.8-stretch  in docker hub

magajh:
@andrey-canon are you going to add the tests for Django30,31 in a different PR?

.circleci/config.yml
```diff
@@ -14,11 +14,14 @@ workflows:
               only: /v?[0-9]+(\.[0-9]+)*/
           matrix:
             parameters:
-              python_version: ["2.7", "3.5"]
-              django_version: ["django111", "django22"]
```
magajh, line 18: @andrey-canon are you going to add the tests for Django30,31 in a different PR?

andrey-canon: not, that is out of the scope

setup.py
```diff
@@ -37,14 +39,13 @@ def get_version():
     packages=['eox_tenant'],
     classifiers=[
         'Development Status :: 5 - Production/Stable',
-        'Framework :: Django :: 1.11',
+        'Framework :: Django :: 2.20',
```
mariajgrimaldi, line 42: why not 2.2?

MoisesGSalas (approved):

PR #124, opened 2021-05-05
feat!: Add lilac backends.

andrey-canon:
## Description
Add lilac backends, to run in a previous version the settings should be overridden

### DIFF
--- a/eox_tenant/settings/common.py
+++ b/eox_tenant/settings/common.py
@@ -36,12 +36,12 @@ def plugin_settings(settings):
     # Plugin settings.
     settings.EOX_TENANT_CACHE_KEY_TIMEOUT = 300
     settings.GET_OAUTH_DISPATCH_BACKEND = 'eox_tenant.edxapp_wrapper.backends.oauth_dispatch_j_v1'
-    settings.GET_BRANDING_API = 'eox_tenant.edxapp_wrapper.backends.branding_api_h_v1'
+    settings.GET_BRANDING_API = 'eox_tenant.edxapp_wrapper.backends.branding_api_l_v1'
     settings.GET_CERTIFICATES_MODULE = 'eox_tenant.edxapp_wrapper.backends.certificates_module_i_v1'
     settings.GET_SITE_CONFIGURATION_MODULE = 'eox_tenant.edxapp_wrapper.backends.site_configuration_module_i_v1'
     settings.GET_THEMING_HELPERS = 'eox_tenant.edxapp_wrapper.backends.theming_helpers_h_v1'
     settings.EOX_TENANT_EDX_AUTH_BACKEND = "eox_tenant.edxapp_wrapper.backends.edx_auth_i_v1"
-    settings.EOX_TENANT_USERS_BACKEND = 'eox_tenant.edxapp_wrapper.backends.users_i_v1'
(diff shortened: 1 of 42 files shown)


mariajgrimaldi (approved):

mariajgrimaldi (approved):

MoisesGSalas (approved):

jignaciopm (approved):

magajh (approved):

eox_tenant/settings/common.py
```diff
@@ -36,12 +36,12 @@ def plugin_settings(settings):
     # Plugin settings.
```
magajh, line 36: @andrey-canon there is a discussion in this [PR](https://github.com/eduNEXT/eox-core/pull/143) about not changing the defaults yet and wait for the official release of lilac. Maybe we should come to an agreement on this for the plugins

@mariajgrimaldi @morenol

eox_tenant/settings/common.py
```diff
@@ -36,12 +36,12 @@ def plugin_settings(settings):
...
     settings.EOX_TENANT_CACHE_KEY_TIMEOUT = 300
     settings.GET_OAUTH_DISPATCH_BACKEND = 'eox_tenant.edxapp_wrapper.backends.oauth_dispatch_j_v1'
-    settings.GET_BRANDING_API = 'eox_tenant.edxapp_wrapper.backends.branding_api_h_v1'
+    settings.GET_BRANDING_API = 'eox_tenant.edxapp_wrapper.backends.branding_api_l_v1'
     settings.GET_CERTIFICATES_MODULE = 'eox_tenant.edxapp_wrapper.backends.certificates_module_i_v1'
     settings.GET_SITE_CONFIGURATION_MODULE = 'eox_tenant.edxapp_wrapper.backends.site_configuration_module_i_v1'
     settings.GET_THEMING_HELPERS = 'eox_tenant.edxapp_wrapper.backends.theming_helpers_h_v1'
     settings.EOX_TENANT_EDX_AUTH_BACKEND = "eox_tenant.edxapp_wrapper.backends.edx_auth_i_v1"
-    settings.EOX_TENANT_USERS_BACKEND = 'eox_tenant.edxapp_wrapper.backends.users_i_v1'
+    settings.EOX_TENANT_USERS_BACKEND = 'eox_tenant.edxapp_wrapper.backends.users_l_v1'
```
magajh, line 44: @andrey-canon there is a discussion in this [PR](https://github.com/eduNEXT/eox-core/pull/143) about not changing the defaults yet and wait for the official release of lilac. Maybe we should come to an agreement on this for the plugins

@mariajgrimaldi @morenol

andrey-canon: Waiting for the official release doesn't make sense for me since those changes were included in Koa, so I can change the pr name to `Add koa backends `and the setting name to `users_k_v1` on the other hand as @morenol  mentioned `For the SaaS we can define EOX_CORE_USERS_BACKEND with another value but that is also something that can be done by someone that wants to use this with Lilac. ` the discussion here is who will have to set new values, the user who is using  last versions(platform and plugin) or the user who wants to use the last version of the plugin with an old edx-platform version. As a developer in a public repo I prefer as less configuration as possible in the last release if a user wants to use the last version could  update the platform or  add the extra configuration or why the user who is in the last version should do more work ?

PR #137, opened 2021-11-30
Return only a serializable subset of settings in `site_values`

MoisesGSalas:
## Description
Some [recent changes](https://github.com/edx/edx-platform/pull/26435) use the value of `site_values` in the email templates context, this causes errors when attempting to serialize the email message because eox-tenant returns a dictionary with all the Django Settings values. To avoid this, only return a serializable subset of the settings.

## Testing instructions

Try to send an email (one of the easiest is account verification)
- Create a new user
- Set the user as inactive
- Try to login

The email should be sent without errors and with the correct context for the tenant.

## Additional information
I performed a small grep of `site_values` on edunext-platform and ignoring the test files and po files it's mostly used to obtain the value of `course_org_filter` and `COURSE_CATALOG_API_URL`.

Considering our current pattern is to access the modified settings object directly (usually via `getattr(setting, "MY_SETTING", default)` instead of interfacing through `SiteConfiguration` (this is mostly done by upstream, and the original `site_values` is a subset of what we are returning here in the proxy`) I think this change is safe.

## Checklist for Merge

- [ ] Updated documentation
- [x] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_tenant/utils.py
+++ b/eox_tenant/utils.py
@@ -19,6 +19,41 @@ domain_pattern = re.compile(
     r'+[A-Za-z0-9][A-Za-z0-9-_]{0,61}'  # First 61 characters of the gTLD
     r'[A-Za-z]$'  # Last character of the gTLD
 )
+NON_SERIALIZABLE = object()
+
+
+def clean_serializable_values(obj):
+    """
+    Returns only serializable values
+
+    If an object is not json serializable returns NON_SERIALIZABLE.
+    In case of a list or dict ignore the NON_SERIALIZABLE values.
+    """
+    if isinstance(obj, (str, int, float, float, bool, type(None))):
+        return obj
+
+    if isinstance(obj, list):
+        return [
+            cleaned_item
+            for cleaned_item in (clean_serializable_values(item) for item in obj)
+            if cleaned_item is not NON_SERIALIZABLE
+        ]
+
+    if isinstance(obj, tuple):
+        return tuple(
+            cleaned_item
+            for cleaned_item in (clean_serializable_values(item) for item in obj)
+            if cleaned_item is not NON_SERIALIZABLE
+        )
+
+    if isinstance(obj, dict):
+        return {
+            key: cleaned_value
+            for key, cleaned_value in ((k, clean_serializable_values(v)) for k, v in obj.items())
+            if cleaned_value is not NON_SERIALIZABLE
+        }
+
+    return NON_SERIALIZABLE
 
 
 def is_valid_domain(domain):
--- a/eox_tenant/tenant_wise/proxies.py
+++ b/eox_tenant/tenant_wise/proxies.py
@@ -15,6 +15,7 @@ from eox_tenant.edxapp_wrapper.site_configuration_module import get_site_configu
 from eox_tenant.models import Microsite, TenantConfig, TenantOrganization
 from eox_tenant.organizations import get_organizations
 from eox_tenant.tenant_wise.context_managers import proxy_regression
+from eox_tenant.utils import clean_serializable_values
 
 SiteConfigurationModels = get_site_configuration_models()
 CertificatesModels = get_certificates_models()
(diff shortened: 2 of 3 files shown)


MaferMazu (approved):

felipemontoya:
@MoisesGSalas do you have an idea of the time it takes to return the whole site_values? You could run this on your local instance a few thousand times and meassure the time to calculate.

In terms of the approach, I think this works and will prevent us having the problems we found in the future or even in future releases. Now in terms of the PR, we need to resolve any review comments add some tests to this before merging.

eox_tenant/utils.py
```diff
@@ -19,6 +19,34 @@
...
+            for cleaned_item in (clean_serializable_values(item) for item in obj)
+            if cleaned_item is not NON_SERIALIZABLE
+        ]
+
+    if isinstance(obj, dict):
+        return {
+            key: cleaned_value
+            for key, cleaned_value in ((k, clean_serializable_values(v)) for k, v in obj.items())
+            if cleaned_value is not NON_SERIALIZABLE
+        }
```
mariajgrimaldi, line 47: and tuples?

MoisesGSalas: What do you mean?

mariajgrimaldi: we don't need to consider `isinstance(obj, tuples)`?

eox_tenant/utils.py
```diff
@@ -19,6 +19,34 @@
...
+
+    return NON_SERIALIZABLE
```
felipemontoya, line 49: What happens when you return NON_SERIALIZABLE?
Will this fail the celery serialization? or is this a placeholder to signify that there was a key with some value, but the value would not be serialized and therefore is empty?

eox_tenant/utils.py
felipemontoya: When we add testing we will notice some more of this.

eox_tenant/utils.py
MoisesGSalas: Yes, I use `NON_SERIALIZABLE` as a sentinel to mark values that are not json serializable. When returning from the recursive call the non serializable values get cleaned from the nested data structures (`list`, `dict`). I was going to use `None` but `None` corresponds to `null` and I didn't want to delete keys with those values.

If you were to run something like.

```python
c = {
    "String": "str",
    "integer": 1,
    "bool": True,
    "none": None,
    "object": Exception,
    "list": [1, Exception],
    "dict": {1: "str", "object": Exception},
}
print(clean_serializable_values(c))
```

You would get:

```python
{
    'String': 'str',
    'integer': 1,
    'bool': True,
    'none': None,
    'list': [1],
    'dict': {1: 'str'}
}
```

MoisesGSalas:
@felipemontoya, I did a few runs, and we are taking a considerable hit in performance:

For `N=1000`

- original site_values from the proxy = `0.0031391930097015575`
- This PR = `2.3681439200008754`
- `json.loads(json.dumps(self.values, lambda o: "default"))` = `2.115462176006986`

felipemontoya:
2 whole seconds for this function alone is deep into the unacceptable range. How big is the resulting dictionary? How many keys are we talking about?

MoisesGSalas:
Around 907 keys, 204 of those are either list, dict or tuple, and the sum of the lens of this nested structures (only on the second level) add up to 1428. And doing a quick check seems like the depth of the original dict is around 7 levels.

felipemontoya:
I'm warming up to the idea of using the content of `TenantConfig`.
What would be the cost of the query if we compare it to the other experiment you did before?

MoisesGSalas:
@felipemontoya, I added some basic tests, and caching `site_values` during the life of the object dramatically improves the performance in my (*very flawed*) benchmark (2s->0.06s for 1000 calls)

felipemontoya (approved):

PR #148, opened 2022-08-16
feat: add extra lookup_field to Microsite and TenantConfig viewsets

magajh:
## Description

Currently, the eox-tenant API only allows getting a specific Microsite or TenantConfig through the primary key, which is the autogenerated id.

```
<domain>/eox-tenant/api/v1/microsites/1/
<domain>/eox-tenant/api/v1/configs/9/
```

In this PR, we implement the **AlternativeFieldLookupMixin** class and add the `alternative_lookup_field` attribute for the Microsite and TenantConfig viewsets. 

The mixin allows getting the corresponding object via the autogenerated `id` field (pk) or the field specified in the `alternative_lookup_field` attribute, so now the requests to a specific Microsite can be made both using the id (pk) or the `key` field.  Same for the TenantConfigs, a request over a specific instance can be made with the id or the `external_key` field.

All of the following requests work now
```

<domain>/eox-tenant/api/v1/microsites/xAer5z6FEbW/ (using the microsite Key)
<domain>/eox-tenant/api/v1/microsites/1002/ (using the microsite Id)

<domain>/eox-tenant/api/v1/configs/Ac397kBmd47/ (using the tenant config External Key)
<domain>/eox-tenant/api/v1/configs/332/ (using the tenant config Id)

```

We also improved the docstring of both the Microsites and TenantConfigs Viewsets, so it's easier to know how to make the requests and what to expect in the response.

### More details on the implementation
If the value sent in the URL is an identifier composed of digits only, then the `lookup_field `will remain the same. Otherwise, the `alternative_lookup_field` will become the lookup field for the object. 

The **AlternativeFieldLookupMixin** class also assumes that the `lookup_field` attribute of the view is a field whose value is composed of digits only and expects the `alternative_lookup_field` value to contain non-numeric characters.

### Why are these changes necessary?

The Xman service needs to communicate with the tenant-API to perform operations on a remote Site in edxapp, specifically, to update its configuration values. However, the [Tenant model](https://github.com/eduNEXT/xman/blob/master/apps/tenant_comm_manager/models.py#L302) in Xman only stores the key/external_key value from a Site (whether is a Microsite or a Tenant Config) in the key field. 

By applying these changes we are making sure that the Xman service will be able to communicate with the tenant-API and make external requests to perform operations over a specific Microsite or Tenant Config with the key value stored in its Tenant model.

## Testing instructions

- You can test this by making the following requests to the eox-tenant API
```

PATCH <domain>/eox-tenant/api/v1/microsites/{id}/
PATCH <domain>/eox-tenant/api/v1/microsites/{key}/

PATCH <domain>/eox-tenant/api/v1/configs/{id}/
PATCH <domain>/eox-tenant/api/v1/configs/{external_key}/

```
Try the same cases but with the other methods: DELETE, GET, and PUT

- The improvements in the docstring can be seen in the following links:
```
<domain>/eox-tenant/api/v1/microsites/
<domain>/eox-tenant/api/v1/configs/
```

## Checklist for Merge

- [ ] Updated documentation
- [x] Rebased master/main
- [x] Squashed commits


felipemontoya (changes requested):

eox_tenant/api/v1/viewsets.py
```diff
@@ -9,21 +9,438 @@
...
+        lookup_value = self.kwargs.get(main_lookup_field)
+
+        if lookup_value and not lookup_value.isdigit():
+            alternative_lookup_field = getattr(self, 'alternative_lookup_field', main_lookup_field)
+
+            self.kwargs.pop(main_lookup_field, None)
+
+            self.kwargs[alternative_lookup_field] = lookup_value
+
+            self.lookup_field = alternative_lookup_field
```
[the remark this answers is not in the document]
magajh: @felipemontoya thanks for the review. I added a warning for the case where the id is used

Henrrypg (approved):

JuanDavidBuitrago (dismissed):

eox_tenant/api/v1/viewsets.py
```diff
@@ -1,29 +1,451 @@
...
+    This mixin allows to get the corresponding object via the autogenerated id field (pk)
+    or the field specified in the 'alternative_lookup_field' attribute.
+
+    If the value sent in the URL is an identifier composed of digits only, then the
+    lookup_field will remain the same. Otherwise, the alternative_lookup_field will become
+    the lookup field for the object.
+
+    This class assumes that the 'lookup_field' attribute of the view is a field whose value
+    is composed of digits only and expects the 'alternative_lookup_field' value to contain
+    non-numeric characters.
```
felipemontoya, line 27: Since this deprecates the use of the regular ID I would suggest that we leave a comment here stating that this Mixin can be removed after some date.
I would put July 2023 as tentative. After that, then we can remove this and only use the alternative.

eox_tenant/api/v1/viewsets.py
```diff
@@ -1,29 +1,451 @@
...
+            ]
+        }
+
+    **Example Update Request**:
+
+    `PATCH <domain>/eox-tenant/api/v1/microsites/xAer5z6FEbW/ (Sending microsite key)`
 
-class MicrositeViewSet(viewsets.ModelViewSet):
-    """MicrositeViewSet that allows the basic API actions."""
+    `PATCH <domain>/eox-tenant/api/v1/microsites/1001/ (Sending microsite id)`
```
felipemontoya, line 215: We can remove the example using the id. We don't want anyone using it. Let's not encourage it in the docs.

eox_tenant/api/v1/viewsets.py
```diff
@@ -1,29 +1,451 @@
...
+                "meta": {},
+                "external_key": "xAer5z6FEbW",
+                "organizations": []
+            },
+            ...
+        ]
+
+    `GET <domain>/eox-tenant/api/v1/configs/xAer5z6FEbW/ (Sending tenant_config external_key)`
+
+    `GET <domain>/eox-tenant/api/v1/configs/207/ (Sending tenant_config id)`
```
felipemontoya, line 310: idem

eox_tenant/api/v1/viewsets.py
```diff
@@ -1,29 +1,451 @@
...
+            "organizations": [
+                2
+            ]
+        }
+
+    **Example Update Request**:
+
+    `PATCH <domain>/eox-tenant/api/v1/configs/xAer5z6FEbW/ (Sending tenant_config external_key)`
+
+    `PATCH <domain>/eox-tenant/api/v1/configs/1001/ (Sending tenant_config id)`
```
felipemontoya, line 391: idem

felipemontoya (approved):
Since we have no insight into anyone that could be using the API with the ID already, I hink this is acceptable.

We should communicate here in the changelogs that this usage is deprecated and remove it when we update. I suggested a date above, but perhaps we can be more agressive and remove it when we update to olive.

felipemontoya (changes requested):
sorry I meant to request changes with the latest round of feedback.

@Henrrypg and @JuanDavidBuitrago did you run it? could you provide any feedback into your revision/tests/process?

JuanDavidBuitrago:
> sorry I meant to request changes with the latest round of feedback.
> 
> @Henrrypg and @JuanDavidBuitrago did you run it? could you provide any feedback into your revision/tests/process?

Hi!! @felipemontoya I ran in ``stack builder`` with ``mango``. I tested with _id_ and _key field_, I made POST, PATCH and GET requests. I just realized that I forgot to test DELETE request.

I'm going to test with DETELE request and let you know the result.

JuanDavidBuitrago:
> 

@felipemontoya and @magajh, I tested DELETE request too, all the endpoint are working with this feature.

Let me know if you need something else.

magajh:
> > 
> 
> @felipemontoya and @magajh, I tested DELETE request too, all the endpoint are working with this feature.
> 
> Let me know if you need something else.



> > 
> 
> @felipemontoya and @magajh, I tested DELETE request too, all the endpoint are working with this feature.
> 
> Let me know if you need something else.

@JuanDavidBuitrago Awesome. Thank you so much !!

eox_tenant/api/v1/viewsets.py
magajh: Done

eox_tenant/api/v1/viewsets.py
magajh: Done

eox_tenant/api/v1/viewsets.py
magajh: Done

eox_tenant/api/v1/viewsets.py
magajh: @felipemontoya done
https://github.com/eduNEXT/eox-tenant/blob/04a47c59db77041252abca6c951b9c5cd236fa61/eox_tenant/api/v1/viewsets.py#L29-L32

felipemontoya (dismissed):
All looking good here

Henrrypg (approved):

PR #149, opened 2022-08-21
feat: add BearerAuthentication to API

magajh:
## Description

Currently, the tenant API permission class [EoxTenantAPIPermission](https://github.com/eduNEXT/eox-tenant/blob/master/eox_tenant/api/v1/permissions.py#L11) only allows superusers to make requests through a basic auth scheme.

https://github.com/eduNEXT/eox-tenant/blob/master/eox_tenant/api/v1/permissions.py#L10-L13

In this PR we add a Bearer authentication scheme to the API

A new file containing the tests for the API permissions was added, and the test cases for the permissions were removed from the API tests.

### More details on the implementation

This implementation was based on the eox-core API permission protocol
https://github.com/eduNEXT/eox-core/blob/master/eox_core/api/v1/permissions.py#L32
 
To grant access, the EoxTenantAPIPermission class now checks multiple things:

- Staff members always have access, so If the requesting user is staff, returns True.
- For a non-staff user: 
  - If there is no valid SignUp source for the user and site from the request, then returns False. 
  - If the user is calling the API from a valid source,  then checks if the domain is authorized by the auth application.allowed_uris or client.URL. 
  -  Finally, if the site is authorized by the Auth, validates that the user has the **can call eox-tenant API** permission

### Why are these changes necessary?

The Xman service needs to communicate with the tenant-API to perform operations on a remote Site in edxapp. However, the protocol in Xman to authenticate in a remote site requires the external API to allow a Bearer Token authentication scheme.
https://github.com/eduNEXT/xman/blob/master/apps/tenant_comm_manager/models.py#L147

By applying these changes we are making sure that the Xman service can safely authenticate to the tenant-API and make external requests to perform operations over a specific Microsite or Tenant Config.

## Testing instructions

STEPS

1. Create an application for the user in `<lms-domain>/admin/oauth2_provider/application/`
2. Add the user to the **EOX Core API access** Group
3. Add a SignUp source instance with the user and the site from where the API calls will be made `<lms-domain>/admin/student/usersignupsource/`
4. Generate a token with the client ID and client secret from the Application `<domain>/oauth2/access_token/`
5. make requests to the eox-tenant API using BearerAuthentication

You can also test using a staff user that does not have a signup source on the request site, and it should grant access to the API

## Checklist for Merge

- [ ] Updated documentation
- [x] Rebased master/main
- [x] Squashed commits

### DIFF
--- a/eox_tenant/api/v1/permissions.py
+++ b/eox_tenant/api/v1/permissions.py
@@ -1,13 +1,66 @@
 """
 Permission for eox_tenant api v1.
 """
-from rest_framework import permissions
+from django.conf import settings
+from django.contrib.auth.models import Permission, User
+from django.contrib.contenttypes.models import ContentType
+from django.db.utils import ProgrammingError
+from rest_framework import exceptions, permissions
+
+
+def load_permissions():
+    """
+    Helper method to load a custom permission on DB that will be
+    used to give access to the eox-tenant API.
+    """
+    if settings.EOX_TENANT_LOAD_PERMISSIONS:
+        try:
+            content_type = ContentType.objects.get_for_model(User)
+            Permission.objects.get_or_create(  # pylint: disable=unused-variable
+                codename='can_call_eox_tenant',
+                name='Can access eox-tenant API',
+                content_type=content_type,
+            )
+        except ProgrammingError:
+            # This code runs when the app is loaded, if a migration has not been
+            # done a ProgrammingError exception is raised.
+            # we are bypassing those cases to let migrations run smoothly.
+            pass
 
 
 class EoxTenantAPIPermission(permissions.BasePermission):
-    """Only allows super user."""
+    """
+    Defines a custom permissions to access eox-tenant API.
+    These permissions make sure that a token is created with the client credentials of the same site is
+    being used on, and the user has a valid SignUp source for the site.
+    """
 
     def has_permission(self, request, view):
-        if request.user.is_superuser:
+        """
+        To grant access, checks if the requesting user:
+            1) it's a staff user
+            3) it's calling the API from a site authorized by the auth application or client
+            4) has can call eox-tenant API permission
+        """
+        user = request.user
+
+        if user.is_staff:
             return True
-        return False
+
+        try:
+            application_uri_allowed = request.auth.application.redirect_uri_allowed(request.build_absolute_uri('/'))
+        except Exception:  # pylint: disable=broad-except
+            application_uri_allowed = False
+
+        try:
+            client_url_allowed = request.get_host() in request.auth.client.url
+        except Exception:  # pylint: disable=broad-except
+            client_url_allowed = False
+
+        if client_url_allowed or application_uri_allowed:
+            return user.has_perm('auth.can_call_eox_tenant')
+
+        # If we get here either someone is using a token created on one site in a different site
+        # or there was a missconfiguration of the oauth client.
+        # we return the most basic message to prevent leaking important information .
+        raise exceptions.NotAuthenticated(detail="Invalid token")
--- a/eox_tenant/api/v1/tests/test_microsites.py
+++ b/eox_tenant/api/v1/tests/test_microsites.py
@@ -129,9 +129,3 @@ class MicrositeAPITest(APITestCase):
         response = self.client.delete(self.url_detail)
 
         self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
-
-    def test_permissions_microsite(self):
-        """Must return 403, only allows superuser."""
-        response = self.client.get(self.url)
-
-        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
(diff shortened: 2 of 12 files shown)


eox_tenant/api/v1/permissions.py
```diff
@@ -1,13 +1,88 @@
...
+    @staticmethod
+    def is_authorized_source_for_user(request_host, user_signup_sources):
+        """
+        Returns True if the current user has a valid signUp source to make requests
+        to the current site.
+        """
+        authorized_sources = getattr(settings, 'EDNX_ACCOUNT_REGISTRATION_SOURCES', [request_host])
+
+        for source in user_signup_sources:
+            if any(re.match('{0}$'.format(pattern), source.site, re.IGNORECASE) for pattern in authorized_sources):
```
magajh, line 85: I'm using format instead of f-string to keep the compatibility with Juniper

JuanDavidBuitrago (dismissed):

Henrrypg (dismissed):

eox_tenant/api/v1/permissions.py
```diff
@@ -1,13 +1,88 @@
...
+        # we return the most basic message to prevent leaking important information .
+        raise exceptions.NotAuthenticated(detail="Invalid token")
+
+    @staticmethod
+    def is_authorized_source_for_user(request_host, user_signup_sources):
+        """
+        Returns True if the current user has a valid signUp source to make requests
+        to the current site.
+        """
+        authorized_sources = getattr(settings, 'EDNX_ACCOUNT_REGISTRATION_SOURCES', [request_host])
```
felipemontoya, line 82: If I understand this correctly, when there is no EDNX_ACCOUNT_REGISTRATION_SOURCES at the site, then authorized_sources is the host and thus this check will always pass. Right?

eox_tenant/edxapp_wrapper/backends/bearer_authentication_m_v1.py
```diff
@@ -0,0 +1,16 @@
+"""Backend for authentication.
```
felipemontoya, line 1: Is this backend compatible with Lilac? Could we name it ..._l_v1 so that we create this functionality already expecting that limonero can use it?

felipemontoya (commented):
It looks good just from reading the code.

felipemontoya:
@JuanDavidBuitrago or @Henrrypg did you guys ran this code in the stack?

JuanDavidBuitrago:
> @JuanDavidBuitrago or @Henrrypg did you guys ran this code in the stack?

Hi!!! @felipemontoya, I ran it. I used ``stack builder`` with ``mango``. I tested with a staff user and it allowed to use the API, also I used a normal user when don't have permission to perform and avoid use it.

@felipemontoya or @magajh maybe need to check something else?

eox_tenant/api/v1/permissions.py
felipemontoya: This would be OK for me. But I'm also wondering, should we for this to have a registered source in the site and that is it. We don't need to support the feature EDNX_ACCOUNT_REGISTRATION_SOURCES in here.

The cost of not doing it is that every single site needs its own user to call this API. It does not sound so bad.

What I think is worse is that sites that do not use the `user_signup_sources` feature are completely blocked from this.
Now I get the sense that we should make validating the `user_signup_sources` optional according to a setting and leave the ednx_account... completely out of it.

What are your thoughts on this @Alec4r ?

magajh: @felipemontoya I thought that all the sites were going to use the `user_signup_sources` feature, but if that's not the case, then I think we should remove the signup source validation. Do you think it's really worth it to leave it as an option?

In fact, the eox-core permission protocol doesn't make this signup source validation, it's something extra that I added because I thought it made sense https://github.com/eduNEXT/eox-core/blob/master/eox_core/api/v1/permissions.py#L32

magajh:
> > @JuanDavidBuitrago or @Henrrypg did you guys ran this code in the stack?
> 
> Hi!!! @felipemontoya, I ran it. I used `stack builder` with `mango`. I tested with a staff user and it allowed to use the API, also I used a normal user when don't have permission to perform and avoid use it.
> 
> @felipemontoya or @magajh maybe need to check something else?

@JuanDavidBuitrago Nice. Thank you!

eox_tenant/edxapp_wrapper/backends/bearer_authentication_m_v1.py
magajh: Yes you are right, it is compatible with Lilac
Thanks @felipemontoya, I changed the file name

magajh: actually, It's compatible with Juniper as well

eox_tenant/api/v1/permissions.py
felipemontoya: This makes sense but only in the case of a large multi-tenant installation such as our saas service. I would not want to make everyone have to fight this restriction for something that is pure edunext service.

magajh: @felipemontoya I removed the SignUp source verification

felipemontoya (approved):
This is looking good for me. Ready to merge

felipemontoya:
@JuanDavidBuitrago could you help us get this merged and release?

JuanDavidBuitrago (approved):

PR #164, opened 2023-01-18
perf(DS-365): olive support

Alec4r:
These commits:
- [PERF] Remove the useless backends for versions older than maple.
- [FEAT] Add compatibility for Python, 3.10 y 3.11.
- [FEAT] Add compatibility for Django 32.
- [PERF] Remove compatibility for django 22.
- [CI] Add the new python and django versions to the matrix test in the github actions workflow.

Nota:
The python and django versions were upgraded following this document: https://docs.google.com/spreadsheets/d/11DheEtMDGrbA9hsUvZ2SEd4Cc8CaC4mAfoV8SVaLBGI/edit#gid=195838733

### How to test
You should have an tutor environment and install eox-tenant as a private package.
Then you should run the test cases defined in this document: https://docs.google.com/document/d/1921hOajxAYrdin_thEZ005aQBwPNoj2zsbgkf3q1Mjc/edit.

### Info
How to add eox-tenant: https://docs.tutor.overhang.io/dev.html#xblock-and-edx-platform-plugin-development
Create user for tutor: https://docs.tutor.overhang.io/local.html#creating-a-new-user-with-staff-and-admin-rights
How to configure multitenant: https://docs.google.com/document/d/1921hOajxAYrdin_thEZ005aQBwPNoj2zsbgkf3q1Mjc/edit#heading=h.doulrbfn1rer


mariajgrimaldi:
are the test cases updated? I thought these changes were removed:
https://github.com/eduNEXT/edunext-platform/pull/438
```
TenantSiteConfigProxy
TenantGeneratedCertificateProxy
```
how can those be tested?

.github/workflows/tests.yml
```diff
@@ -11,8 +11,8 @@ jobs:
     strategy:
       max-parallel: 2
       matrix:
-        python-version: ["3.8"]
-        django: ["22","32"]
+        python-version: ["3.8", "3.10", "3.11"]
```
mariajgrimaldi, line 14: Which criteria you're using here in order to add 3.10 and 3.11? should we adopt it in the other plugins?

.github/workflows/tests.yml
```diff
@@ -11,8 +11,8 @@ jobs:
...
+        django: ["32"]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
```
mariajgrimaldi, line 18: Would you mind updating also the actions versions?

README.rst
```diff
@@ -29,8 +29,10 @@ Compatibility Notes
 +-------------------+-----------------------+
 |      Nutmeg       |         >= 6.2        |
 +-------------------+-----------------------+
+|       Olive       |         >= 8.0        |
```
MaferMazu, line 32: Maybe setting an upper bound to lilac is a good idea because we are still not testing with Django 22. What do you think?

MaferMazu (commented):
The code looks good; I am going to do the test cases.

I left a comment, and I am going to open a question. I don't know if it's good to maintain a lot of constraints; we had it because before, we were far from master and the updates, but now maybe it's a good moment to clean some of the constraints. What do you think?

.github/workflows/tests.yml
Alec4r: @mariajgrimaldi no, we spoke with Felipe about this and the "support windows file" and he explained me how to read the file and 3.10 and 3.11 is not necessary, so we can remove them, however I think that is a good practice fix the deprecations warnings to doesn't have problems in a future.

MaferMazu:
About the test cases:
- Tenant API ✅
- Commands: I really didn't understand very well the output and used a lot of microsite
- Enrollments
- Proxies ✅ (Maybe it's a good idea to edit how to test it because the tenant conf proxy we are testing it when we change the domain and see different information, and about the certificate, we drop it)
- Multitenant ✅
- Info view ✅
- Template tags (I tried editing the lms/templates/main.html, and it's not working, and with wiki, we need to create a [waffle flag](https://github.com/eduNEXT/tutor-contrib-edunext-distro/blob/5477aac7c35d34fa617a3f1dc583660913719a60/tutordistro/templates/distro/tasks/lms/init) for legacy frontend, but it's not working)

mariajgrimaldi (approved):
is there something else to be addressed here?

JuanDavidBuitrago (approved):
I think it could be merging.

PR #166, opened 2023-01-24
feat: Add course_home api multitenant restriction, based on the curre…

andrey-canon:
## Description

Course Home API has the following endpoints

image

Some of them depends on the course id, however if someone tries to access the result will be successful if the course exists, no matter if the course doesn't belong to the current site, so this pr copies current validations of `/courses/<course-id>` to  `/api/course_home/something/,course-id>` in orderr to return  404 if the course doesn't belong to the current site.

## Testing instructions

1. Go to any endpoint, for example, /api/course_home/course_metadata/<your course id>
2. Then go to a different tenant, that tenant must have the setting course_org_filter with a different value from the course org, the result should be a 404 error.

### Example of configuration:

Tenant 1
```
{
    "course_org_filter": "edx"
}
```
Tenant 2
```
{
    "course_org_filter": "test"
}
```
if  course_id is course-v1:edx+CS101+2022_T4 the api result will be visible in tenant 1 and tenant 2 will return 404.

### DIFF
--- a/eox_tenant/middleware.py
+++ b/eox_tenant/middleware.py
@@ -40,7 +40,10 @@ class MicrositeCrossBrandingFilterMiddleware(MiddlewareMixin):
         microsite, but it is not the current microsite
         """
         path = request.path_info
-        regex_path_match = re.compile('/courses/{}/'.format(settings.COURSE_ID_PATTERN))
+        restricted_courses_pattern = "|".join(settings.EOX_TENANT_RESTRICTED_COURSE_PATTERNS)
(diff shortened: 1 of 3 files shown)


johanv26 (approved):
I approve the implementation of the list to manage block_path with regex. also liked the ddt implementation with list behavior. I only have one question: how or where can I check that each test ran 8 times for `COMMON_COURSE_PATHS` ?

johanv26:
probably the way is looking the increase of test run.

PR #190, opened 2023-11-17
perf: added palm support and changed default settings to always use e…

Asespinel:
This PR mainly updates the requirements for Palm support, and it modifies the setting USE_EOX_TENANT = True as I beleive we should have this as a default if eox_tenant is being installed and ready to use.

## Testing instructions

1. Install this version of eox-tenant in a palm environment.
2. Create routes and tenant configs accordingly as a super user
3. For example:
Screenshot from 2023-11-16 20-36-29
4. If you go to the specified route you should read the "site1" platform name.
5. Other tests cases can be found at: https://docs.google.com/document/d/1921hOajxAYrdin_thEZ005aQBwPNoj2zsbgkf3q1Mjc/edit


bra-i-am (dismissed):
Working as expected!

MaferMazu (commented):
Thanks for this PR, @Asespinel.
When we create support for a new release, we usually use a perf commit instead of a feat because we typically change many things. And for this PR, we are changing a super important variable default value; it is a breaking change. In both cases (new release support and breaking changes), we want to upgrade the version to a major release (for example, if we are in v9.X.X, upgrade to v10.X.X); I suggest we use perf in the PR name.

Reviewing other things:

- [x]  Installation. (Works as expected)
- [x] Backends.
- [x] Github actions.
- [ ] Test cases.
- [ ] Improvements. (Are we going to support JWT tokens? Or would you like us to do that later? I think it is important because, with that improvement, we are already preparing for the [Bearer deprecation](https://github.com/openedx/edx-drf-extensions/issues/284), but if we don't want to do it now, we need to add it in the issues).
- [ ] Requirements. (Are we sure we want to maintain the current constraints? We have openedx_filters==0.8.0, but the version in [edx-platform for palm is 1.2.0](https://github.com/openedx/edx-platform/blob/open-release/palm.master/requirements/edx/base.txt#L775), the same with [mock](https://github.com/openedx/edx-platform/blob/open-release/palm.master/requirements/edx/base.txt#L715); here I suggest we remove the unnecessary constraints and use it only if we need to be restrictive with some package, for example, the `Django` and the `pymongo` constraints are perfect because follow the [edx-platform constraints](https://github.com/openedx/edx-platform/blob/open-release/palm.master/requirements/constraints.txt), another thing can help use constraints but not too restrictive with ==, for example, maybe we want to constraint pylint to be < to 3.0.0, but no force to be a specific version to maintain congruency with [edx-platform pylint version](https://github.com/openedx/edx-platform/blob/open-release/palm.master/requirements/constraints.txt))
- [ ] Documentation. (Taking into account the first comment, the right version to support palm in the readme is >=v10.0.0, and it will be helpful if you can update the drive documentation resolving the comments)

Asespinel:
Hi @MaferMazu Thanks for the comments they are always greatly appreciated.

**Improvements**
I left an Issue regarding the JWT that we can tackle in our next sprints.

**Requirements**
I revisited the constraints file and changed it accordingly. Please let em know if you have additional comments so I can modify the file accordingly

**Documentation**
I changed the readme witht he right version of eox tenant and changed the init file as well. Regarding the drive documentation, Im going to take another look to the changes and I'll work on it.

I also modified the PR title to better suit the objective of the PR

dcoa (approved):

dcoa (approved):

MaferMazu (approved):
Thanks for everything, @Asespinel. It looks good to me. I suggest avoiding editing the version in the init because the bumpversion workflow changes it automatically. Following #164, it is better to avoid touching that file. @Alec4r that's the only comment I have; you can merge!

Alec4r (commented):
revert: update version

Alec4r (approved):

MaferMazu (approved):

PR #192, opened 2023-12-19
refactor: inherit from social auth exception

johanseto:
## Description

This changes the inheritance of the eox-tenant-auth-exception.

This is with the purpose of the exception to be handled by the social tpa middleware exception process.
This is to avoid this 502 error msg:
image
image

You can see here that the [new inheritance  class](https://github.com/python-social-auth/social-core/blob/29cbbd22b98d81d569a886b1cc0bd9a316cd124f/social_core/exceptions.py#L1) is also based on value error exception. 

**But the change is related to this[ eox-core PR](https://github.com/eduNEXT/eox-core/pull/171):**

So now the exception is family of the SocialAuthBaseException.

With that change now, the [social_django middleware](https://github.com/python-social-auth/social-app-django/blob/5.4.0/social_django/middleware.py#L35) could manage it.

Keep in mind that this middleware is the parent of[ edx-platform middleware](https://github.com/openedx/edx-platform/blob/ebcbe1cd9208191c0589d7fe538c6ac13470abe6/common/djangoapps/third_party_auth/middleware.py#L18). 

The middleware seems applied in the [openedx tpa module](https://github.com/openedx/edx-platform/blob/ebcbe1cd9208191c0589d7fe538c6ac13470abe6/common/djangoapps/third_party_auth/settings.py#L27).

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits


andrey-canon (commented):
Why you are not adding this to the requirements ?
Tests failing

johanseto:
For the test failing I made this PR to eox-tenat. But yes I can added also to this branch in other PR.
https://github.com/eduNEXT/eox-tenant/pull/193

johanseto:
@andrey-canon for the inclusion of the requirements. I avoid it because I want this PR to have the lowest lines and lowest impact.
Also, I want this to be moved to master, so then I would have a requirements incompatibility due to the large update...
Also some other eox`s like core have block code like that:
https://github.com/eduNEXT/eox-core/blob/master/eox_core/middleware.py#L31-L37
Also this eox:
https://github.com/eduNEXT/eox-tenant/blob/master/setup.py#L7-L10

andrey-canon:
> @andrey-canon for the inclusion of the requirements. I avoid it because I want this PR to have the lowest lines and lowest impact. Also, I want this to be moved to master, so then I would have a requirements incompatibility due to the large update... Also some other eox`s like core have block code like that: https://github.com/eduNEXT/eox-core/blob/master/eox_core/middleware.py#L31-L37 Also this eox: https://github.com/eduNEXT/eox-tenant/blob/master/setup.py#L7-L10

in the past this was consider as the lazy way to pass the test if you are worried about the number of lines my suggestion is to do `make upgrade`  and then discard all the changes that are not related with the social-core dependency

johanseto:
> 4961cef

[4961cef](https://github.com/eduNEXT/eox-tenant/pull/192/commits/4961cef952b069345d90f3064831d7a219637505)

johanseto:
> in the past this was consider as the lazy way to pass the test if you are worried about the number of lines my suggestion is to do `make upgrade` and then discard all the changes that are not related with the social-core dependency

[4961cef](https://github.com/eduNEXT/eox-tenant/pull/192/commits/4961cef952b069345d90f3064831d7a219637505)

johanseto:
@andrey-canon I have added the dependency.
[b306762](https://github.com/eduNEXT/eox-tenant/pull/192/commits/b306762481fbf831e3730862851d5a4c79cb1620)

andrey-canon:
@johanseto could you open the PR against master, please

andrey-canon (approved):

johanseto:
PR opened against master
https://github.com/eduNEXT/eox-tenant/pull/195

PR #197, opened 2024-02-22
fix: allow using boolean as values in tenant configs when getting settings by organization

bra-i-am:
## Description

This PR is created for **a necessity arisen in the [PR #808 from edunext-platform](https://github.com/eduNEXT/edunext-platform/pull/808)** and allows using boolean types as values because when some tenant properties have a value `False` or `True`, this is used to run two conditionals. e.g. if a tenant setting `ENABLE_COURSE_AUTHORING_MFE` has a value `False`, even though we want to send the value `False`, the conditional is skipped and is returned a None instead of `False` 

## Testing instructions

At this moment the useful way to test this PR would be running the [PR #808 from edunext-platform](https://github.com/eduNEXT/edunext-platform/pull/808)

### DIFF
--- a/eox_tenant/models.py
+++ b/eox_tenant/models.py
@@ -216,7 +216,7 @@ class TenantConfig(models.Model):
         for result in results:
             value = result.lms_configs.get(val_name)
 
-            if value:
+            if value is not None:
                 return value
 
         return None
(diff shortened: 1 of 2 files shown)


MaferMazu (approved):
This looks good to me ✨
Please rebase to ensure the quality test is still passing!

PR #198, opened 2024-03-05
feat: add route domain to proxy the site `domain`

andrey-canon:
## Description

This add the route domain if the tenant has tenant route a associated. This also avoid the creation of a site record with the same domain of a route created.

Migration pr of  https://github.com/eduNEXT/eox-tenant/pull/162


johanseto (approved):
LGTM. May be we could move this to eox-tenant master as a candidate

PR #203, opened 2024-05-06
feat: allow user set language preference

dcoa:
This PR has the intention to allow users to change the website language preference based on their cockies. 

This PR is inspired by https://github.com/eduNEXT/eox-nelp/pull/127 and allows the normal behavior of the header and footer lang selectors for Django templates.


dcoa:
Closed in favor of https://github.com/eduNEXT/eox-core/pull/265

PR #204, opened 2024-05-16
feat: add integration test

luisfelipec95:
## Description
This PR add integration test file

## Testing instructions

- In a Quince or Palm environment install eox-tenant, I used docker-compose.override.yml
- Go to bash:
`tutor dev run lms bash`
`make test-requirements`
`pytest -s --ds=lms.envs.tutor.test /openedx/eox-tenant/eox_tenant/test/tutor`

## Description

https://edunext.atlassian.net/browse/DS-906

### DIFF
--- a/eox_tenant/settings/test.py
+++ b/eox_tenant/settings/test.py
@@ -3,6 +3,11 @@ Common settings for eox_tenant project.
 """
 from __future__ import absolute_import, unicode_literals
 
+import codecs
+import os
+
+import yaml
+
 from .common import *  # pylint: disable=wildcard-import,unused-wildcard-import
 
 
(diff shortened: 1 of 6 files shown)


Asespinel (approved):
I only got some warnings when running up the tests, however it doesn't seem to affect the tests themselves.

bra-i-am (approved):

MaferMazu:
I have the same comment as in eox-hooks. It would be great if you could add an API test if you think it could enter the scope; if not, we can go with this because it works.

Thanks for this @luisfelipec95 🙌

luisfelipec95:
> I have the same comment as in eox-hooks. It would be great if you could add an API test if you think it could enter the scope; if not, we can go with this because it works.
> 
> Thanks for this @luisfelipec95 🙌

Yes, I think that test is important, it will be done with the integration of GitHub Actions with all the plugins. Thanks

PR #207, opened 2024-05-31
feat: add github action integration

luisfelipec95:
## Description
This PR adds the configuration files to integrate github action https://github.com/eduNEXT/integration-test-in-tutor

## How to test
Checks:

- Integration / integration-test (<17.0.0) (pull_request)

- Integration / integration-test (<18.0.0) (pull_request)

## Other information

https://edunext.atlassian.net/browse/DS-963


.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,18 @@
+name: Integration
+on: [pull_request]
+
+jobs:
+  integration-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        tutor_version: ["<17.0.0", "<18.0.0"]
```
magajh, line 9: Thanks @luisfelipec95! Is there a specific reason we need to have a job for "<17.0.0" as well? Wouldn't it be enough to just use "<18.0.0"?

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,18 @@
+name: Integration
```
magajh, line 1: Since this name will be shown in the checks section of the pull request, I think we should name it something more specific, like "Tutor Integration Tests". I'm suggesting this assuming we will only use this workflow for integration tests with Tutor (which I think would be the best approach for better managing and organization)

.github/workflows/integration-test.yml
magajh, line 9: Is there a specific reason we need to have a job for "<17.0.0" as well? Wouldn't it be enough to just use "<18.0.0"?

eox_tenant/test/tutor/integration.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Install the package in the edxapp env
```
magajh, line 3: This is merely a suggestion, but I'm not sure if the inline comments on lines 3, 7, and 11 add much value. The script is easy to understand, so perhaps a single inline comment at the beginning briefly explaining what the script does would be enough.

magajh (commented):
Thanks for the work @luisfelipec95!

.github/workflows/integration-test.yml
luisfelipec95: > Is there a specific reason we need to have a job for "<17.0.0" as well? Wouldn't it be enough to just use "<18.0.0"?

Because only in Quince and Palm versions can you use the mount command

eox_tenant/test/tutor/integration.sh
luisfelipec95: thank you! corrected in all PR's

.github/workflows/integration-test.yml
luisfelipec95: thank you! corrected in all PR's

eox_tenant/test/tutor/integration_test_tutor.py
BryanttV, line 1: I'm not sure but I think `edx_auth_n_v1` backend should be imported in `test_current_settings_code_imports`

luisfelipec95: You are right, the changes have already been made, thank you

BryanttV (approved):

PR #208, opened 2024-05-31
chore: remove unused backends `DS-960`

BryanttV:
## Description
This PR removes unused backends.

## Testing instructions

Using **Tutor**:

1. Install the plugin with these changes.
2. Test the functionality of the plugin. Everything should work correctly,

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits


mariajgrimaldi:
We need to update the compatibility notes to remove the deleted backend in the ironwood notes: https://github.com/eduNEXT/eox-tenant/blob/master/README.rst#compatibility-notes. Also, could you improve the formatting in the compatibility notes table? Thanks!

BryanttV:
Hi @mariajgrimaldi, thanks for the review.

> We need to update the compatibility notes to remove the deleted backend in the ironwood notes

I also thought about doing it, however, @bra-i-am tells me it should not be removed because this backend does exist if we install a version of `eox-tenant` compatible with Ironwood, e.g. `v2.6.0`. If you check the other backends mentioned in the Ironwood notes, you can see that these files no longer exist but are still in **README**. 

> could you improve the formatting in the compatibility notes table?

Yes, I can format it correctly.

mariajgrimaldi:
@BryanttV: good to know, thanks for the explanation!

bra-i-am (approved):
LGTM!

-----------------------------------------

> We need to update the compatibility notes to remove the deleted backend in the ironwood notes

@mariajgrimaldi you are right, we need to update not only the compatibility notes but also the whole readme so it is more appropriate and trustable, nevertheless, taking into account that the readme at this moment works as @BryanttV mentioned, I believe we can wait for [this card](https://edunext.atlassian.net/browse/DS-971), but if you consider this needs to be done right now, there's no problem to me

BryanttV:
Since another card already exists, I will merge with these changes.

PR #210, opened 2024-06-11
docs: improve the documentation

dcoa:
This PR improves the repository documentation for a better understanding of the eox-tenant plugin.

For the plugin description

>  Plugin for managing multiple tenants (organizations) within a single Open edX instance, replacing the microsites and site configuration features with a more robust multi-tenancy model.

**Proposal** 

➡️ Maintain the settings configuration for the current release. 

- **Simplification**: Find the necessary information without being distracted by details that may no longer be relevant (e.g. Ironwood).
- **Relevance**: Users typically work with recent versions of software. Documenting the latest settings ensures that our documentation is pertinent and useful for most users.
- **Maintenance**:  We lower the risk of outdated information and reduce the effort required to keep the documentation current. (e.g., before Redwood we have that the current configuration works for Juniper, that is not true because the latest version at that moment doesn't support that release). 
- **Accuracy**: During the migration process, we are adding the name of the previous release in the documentation even when the settings are the same, which could lead to unnecessary action for the user.

Also, the packages management like (pypi or npm - use the readme and the latest release to explain the documentation)  if you want to install another version you should select it and that will show you the relevant information for that tag (readme for that version). That behavior is another motivation for this change, showing the information relevant to the version that the user wants to use. 

**Additional information**
[JIRA](https://edunext.atlassian.net/browse/DS-971)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,16 +1,49 @@
-
+==========
 EOX tenant
-----------
-Eox-tenant is a multi-tenancy django app for `edx-platform`_. It is build as an `openedx plugin`_ so even as a django app it will auto install in the larger edx-platform core code once installed in the same python environment.
+==========
+|Maintainance Badge| |Test Badge| |PyPI Badge|
 
-The code is written and maintained by `eduNEXT`_ and it is what we use to support our own multi-tenant services. It was initially created as an extension of the `microsites` and `site_configurations` features of the Open edX platform, but it has grown to completely replace them in order to support more robust multi tenancy model.
+.. |Maintainance Badge| image:: https://img.shields.io/badge/Status-Maintained-brightgreen
+   :alt: Maintainance Status
+.. |Test Badge| image:: https://img.shields.io/github/actions/workflow/status/edunext/eox-tenant/.github%2Fworkflows%2Ftests.yml?label=Test
+   :alt: GitHub Actions Workflow Test Status
+.. |PyPI Badge| image:: https://img.shields.io/pypi/v/eox-tenant?label=PyPI
+   :alt: PyPI - Version
+   
+Eox-tenant is a multi-tenancy Django app for `edx-platform`_. It is built as an `openedx plugin`_ so even as a Django app it will autoinstall in the larger edx-platform core code once installed in the same Python environment.
 
-If you are looking for professional development or support with multi tenancy or multi sites in the Open edX platform, you can reach out at sales@edunext.co
+The code is written and maintained by `edunex`_ and we use it to support our multi-tenant services. It was initially created as an extension of the `microsites` and `site_configurations` features of the Open edX platform, however, it has grown to completely replace them to support a more robust multitenancy model.
+
+If you are looking for professional development or support with multitenancy or multi-sites in the Open edX platform, you can reach out to sales@edunext.co
 
 .. _openedx plugin: https://github.com/edx/edx-platform/tree/master/openedx/core/djangoapps/plugins
 .. _edx-platform: https://github.com/edx/edx-platform/
 .. _eduNEXT: https://www.edunext.co
 
+Installation
+============
+
+#. Add this plugin in your Tutor ``config.yml`` with the ``OPENEDX_EXTRA_PIP_REQUIREMENTS`` setting.
+
+   .. code-block:: yaml
+      
+      OPENEDX_EXTRA_PIP_REQUIREMENTS:
+         - eox-tenant=={{version}}
+         
+#. Save the configuration with ``tutor config save``.
+#. Build an open edx image with ``tutor images build openedx``.
+#. Launch your platform with ``tutor local launch``.
+
+Usage
+=====
+Onece your instance is running you can access the Django admin site and locate the ``EDUNEXT OPENEDX MULTITENANCY`` models.
+
+- Microsites
+- Routes
+- Tenant configs
+- Tenant organizations
+
+ 
 Compatibility Notes
 --------------------
 
@@ -38,7 +71,7 @@ Compatibility Notes
 | Redwood          | >= v11.7.0            |
 +------------------+-----------------------+
 
-**NOTE**: Since 6.2 version, eox-tenant does not support Django 2.2
+**NOTE**: Since the 6.2 version, eox-tenant does not support Django 2.2
 
 The following changes to the plugin settings are necessary. If the release you are looking for is
 not listed, then the accumulation of changes from previous releases is enough.
@@ -132,10 +165,9 @@ From version **1.0.0**\ , middlewares **RedirectionsMiddleware** and **PathRedir
    The table corresponding to the Redirection model will not be deleted but it will be discarded from the Django state
 
 Commands
-########
-
+--------
 Synchronize Organizations
-*************************
+^^^^^^^^^^^^^^^^^^^^^^^^^
 This command will synchronize the course_org_filter values in lms_configs(TenantConfig model) or values(Microsite model) with the TenantOrganization registers, if the organization does not exist, it will be created, otherwise it will be add to the organizations model field.
 
 
@@ -146,15 +178,15 @@ This command will synchronize the course_org_filter values in lms_configs(Tenant
   ./manage.py lms synchronize_organizations --model Microsite # only for Microsite
 
 Create/Edit tenant configuration
-********************************
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 `create_or_update_tenant_config` helps to add or edit ``TenantConfig`` and linked ``Routes`` via command line.
 
 .. code-block:: bash
 
-  # this command will create/edit entry in TenantConfig with external_key lacolhost.com and update its JSONField(s) with passed json content.
+  # This command will create/edit an entry in TenantConfig with external_key lacolhost.com and update its JSONField(s) with passed JSON content.
   ./manage.py lms create_or_update_tenant_config --external-key lacolhost.com --config '{"lms_configs": {"PLATFORM_NAME": "Lacolhost"}, "studio_configs": {"PLATFORM_NAME": "Lacolhost"}}' lacolhost.com studio.lacolhost.com preview.lacolhost.com
 
-  # this command will create/edit entry in TenantConfig with external_key lacolhost.com and update its JSONField(s) with passed json config file content.
+  # This command will create/edit an entry in TenantConfig with external_key lacolhost.com and update its JSONField(s) with passed JSON config file content.
   ./manage.py lms create_or_update_tenant_config --external-key lacolhost.com --config-file /tmp/some.json lacolhost.com studio.lacolhost.com preview.lacolhost.com
 
   # Same as above, but it will override configuration instead of updating it.
@@ -163,14 +195,14 @@ Create/Edit tenant configuration
 Caveats
 -------
 
-- SSO that uses the LMS while authenticating does so with server-to-server communication. Therefore, when the `AvailableScreenMiddleware` gets the current domain, it finds that `lms:8000` as in `SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT` which does not exist, then raises 404 exception. In order to avoid this error, set in your LMS settings file:
+- SSO that uses the LMS while authenticating does so with server-to-server communication. Therefore, when the `AvailableScreenMiddleware` gets the current domain, it finds that `lms:8000` as in `SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT` which does not exist, then raises a 404 exception. To avoid this error, set in your LMS settings file:
 
 .. code-block:: python
 
   SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT = SOCIAL_AUTH_EDX_OAUTH2_PUBLIC_URL_ROOT
 
 How to Contribute
------------------
+=================
 
 Contributions are welcome! See our `CONTRIBUTING`_ file for more
 information – it also contains guidelines for how to maintain high code
--- a/README.rst
+++ b/README.rst
@@ -10,15 +10,11 @@ EOX Tenant
 .. |PyPI Badge| image:: https://img.shields.io/pypi/v/eox-tenant?label=PyPI
    :alt: PyPI - Version
    
-Eox-tenant is a multi-tenancy Django app for `edx-platform`_. It is built as an `openedx plugin`_ so even as a Django app it will be auto-installed in the larger edx-platform core code once is installed in the same Python environment.
-
-The code is written and maintained by `edunext`_ and we use it to support our multi-tenant services. It was initially created as an extension of the `microsites` and `site_configurations` features of the Open edX platform, however, it has grown to completely replace them to support a more robust multitenancy model.
+Eox-tenant is a plugin for `Open edX`_, and part of the Edunext Open Extensions (aka EOX), that replaces the microsites and site_configurations features, offering a more robust multi-tenancy model.
 
 If you are looking for professional development or support with multitenancy or multi-sites in the Open edX platform, you can reach out to sales@edunext.co
 
-.. _openedx plugin: https://github.com/openedx/edx-platform/tree/master/openedx/core/djangoapps/plugins
-.. _edx-platform: https://github.com/openedx/edx-platform/
-.. _eduNEXT: https://www.edunext.co
+.. _Open edX: https://github.com/openedx/edx-platform/
 
 Installation
 ============
@@ -31,8 +27,7 @@ Installation
          - eox-tenant=={{version}}
          
 #. Save the configuration with ``tutor config save``.
-#. Build an open edx image with ``tutor images build openedx``.
-#. Launch your platform with ``tutor local launch``.
+#. Build the image and launch your platform with ``tutor local launch``.
 
 Usage
 =====
@@ -45,7 +40,7 @@ Once your instance is running, you can access the Django admin site and locate t
 - **Tenant organizations:** Link each organization with one or multiple tenants.
 
 Add ``EDNX_USE_SIGNAL = True`` in each microsite/tenant that wants to use the plugin. 
- 
+
 Compatibility Notes
 --------------------
 
@@ -73,7 +68,7 @@ Compatibility Notes
 | Redwood          | >= v11.7.0            |
 +------------------+-----------------------+
 
-**NOTE**: Since the 6.2 version, eox-tenant does not support Django 2.2
+⚠️ Since the 6.2 version, eox-tenant does not support Django 2.2
 
 The plugin is configured for the latest release (Quince). The following changes in the plugin settings should be applied to be used for previous releases.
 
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -1,104 +1,138 @@
...
 
-The code is written and maintained by `eduNEXT`_ and it is what we use to support our own multi-tenant services. It was initially created as an extension of the `microsites` and `site_configurations` features of the Open edX platform, but it has grown to completely replace them in order to support more robust multi tenancy model.
+The code is written and maintained by `edunext`_ and we use it to support our multi-tenant services. It was initially created as an extension of the `microsites` and `site_configurations` features of the Open edX platform, however, it has grown to completely replace them to support a more robust multitenancy model.
 
-If you are looking for professional development or support with multi tenancy or multi sites in the Open edX platform, you can reach out at sales@edunext.co
+If you are looking for professional development or support with multitenancy or multi-sites in the Open edX platform, you can reach out to sales@edunext.co
 
 .. _openedx plugin: https://github.com/edx/edx-platform/tree/master/openedx/core/djangoapps/plugins
 .. _edx-platform: https://github.com/edx/edx-platform/
```
BryanttV, line 20:
```suggestion
.. _edx-platform: https://github.com/openedx/edx-platform/
```

README.rst
```diff
@@ -1,104 +1,138 @@
...
 
-The code is written and maintained by `eduNEXT`_ and it is what we use to support our own multi-tenant services. It was initially created as an extension of the `microsites` and `site_configurations` features of the Open edX platform, but it has grown to completely replace them in order to support more robust multi tenancy model.
+The code is written and maintained by `edunext`_ and we use it to support our multi-tenant services. It was initially created as an extension of the `microsites` and `site_configurations` features of the Open edX platform, however, it has grown to completely replace them to support a more robust multitenancy model.
 
-If you are looking for professional development or support with multi tenancy or multi sites in the Open edX platform, you can reach out at sales@edunext.co
+If you are looking for professional development or support with multitenancy or multi-sites in the Open edX platform, you can reach out to sales@edunext.co
 
 .. _openedx plugin: https://github.com/edx/edx-platform/tree/master/openedx/core/djangoapps/plugins
```
BryanttV, line 19: I think this would be the link: https://github.com/openedx/edx-django-utils/tree/master/edx_django_utils/plugins

README.rst
```diff
@@ -121,57 +155,34 @@ From version **1.0.0**\ , middlewares **RedirectionsMiddleware** and **PathRedir
...
 Contributions are welcome! See our `CONTRIBUTING`_ file for more
 information – it also contains guidelines for how to maintain high code
 quality, which will make your contribution more likely to be accepted.
 
 .. _CONTRIBUTING: https://github.com/eduNEXT/eox-tenant/blob/master/CONTRIBUTING.rst
+
+License
+=======
+
+This project is licensed under the AGPL-3.0 License. See the LICENSE file for details.
```
BryanttV, line 188:
```suggestion
This project is licensed under the AGPL-3.0 License. See the `LICENSE.txt <LICENSE.txt>`_ for details.
```

README.rst
mariajgrimaldi: The openedx plugin should also point to opened

README.rst
```diff
@@ -1,104 +1,138 @@
...
+.. |Test Badge| image:: https://img.shields.io/github/actions/workflow/status/edunext/eox-tenant/.github%2Fworkflows%2Ftests.yml?label=Test
+   :alt: GitHub Actions Workflow Test Status
+.. |PyPI Badge| image:: https://img.shields.io/pypi/v/eox-tenant?label=PyPI
+   :alt: PyPI - Version
+   
+Eox-tenant is a multi-tenancy Django app for `edx-platform`_. It is built as an `openedx plugin`_ so even as a Django app it will be auto-installed in the larger edx-platform core code once is installed in the same Python environment.
+
+The code is written and maintained by `edunext`_ and we use it to support our multi-tenant services. It was initially created as an extension of the `microsites` and `site_configurations` features of the Open edX platform, however, it has grown to completely replace them to support a more robust multitenancy model.
+
+If you are looking for professional development or support with multitenancy or multi-sites in the Open edX platform, you can reach out to sales@edunext.co
```
mariajgrimaldi, line 17: Can this be rewritten? It could use an update. Mainly I don't think we need to explain what a django plugin is.

README.rst
```diff
@@ -1,104 +1,138 @@
...
 
-If you are looking for professional development or support with multi tenancy or multi sites in the Open edX platform, you can reach out at sales@edunext.co
+   .. code-block:: yaml
+      
+      OPENEDX_EXTRA_PIP_REQUIREMENTS:
+         - eox-tenant=={{version}}
+         
+#. Save the configuration with ``tutor config save``.
+#. Build an open edx image with ``tutor images build openedx``.
+#. Launch your platform with ``tutor local launch``.
```
mariajgrimaldi, line 35: Doesn't `tutor local launch` already builds the image? Or is it just in dev environments? I'm not sure.

README.rst
```diff
@@ -1,104 +1,138 @@
...
-**NOTE**: Since 6.2 version, eox-tenant does not support Django 2.2
-
-The following changes to the plugin settings are necessary. If the release you are looking for is
-not listed, then the accumulation of changes from previous releases is enough.
-
-**Ironwood**
+- **Microsites:** Store the microsite configuration.
+- **Routes:** Configure the URL for a tenant.
+- **Tenant configs:** Store the configuration for each tenant.
+- **Tenant organizations:** Link each organization with one or multiple tenants.
```
mariajgrimaldi, line 45: what's the difference between microsites and tenant configs? Is the microsites model still in use?

README.rst
```diff
@@ -1,104 +1,138 @@
...
+| Palm             | >= v10.0.0     |
++------------------+----------------+
+| Quince           | >= v11.0.0     |
++------------------+----------------+
+
+**NOTE**: Since the 6.2 version, eox-tenant does not support Django 2.2
+
+The plugin is configured for the latest release (Quince). The following changes in the plugin settings should be applied to be used for previous releases.
+
+If the release you are looking for is not listed, the actual configuration is sufficient or it is incompatible with the current eox-tenant version.
```
mariajgrimaldi, line 78: I thought that if wasn't not listed, it's not supported. In what case does the other scenario happen?

README.rst
```diff
@@ -1,104 +1,138 @@
...
 
-  GET_OAUTH_DISPATCH_BACKEND: 'eox_tenant.edxapp_wrapper.backends.oauth_dispatch_j_v1'
+  EOX_TENANT_EDX_AUTH_BACKEND = "eox_tenant.edxapp_wrapper.backends.edx_auth_i_v1"
+
+Those settings can be changed in ``eox_tenant/settings/common.py`` or, for example, in the instance settings.
 
-**Koa (optional)**
+**NOTE**
 
+For version > 10.0.0 you need to enable eox-tenant adding in the LMS configuration:
```
mariajgrimaldi, line 92:
```suggestion
ℹ️ For version > 10.0.0 you need to enable eox-tenant adding in the LMS configuration:
```

README.rst
```diff
@@ -1,104 +1,138 @@
...
 
 .. code-block:: yaml
 
-  GET_OAUTH_DISPATCH_BACKEND: 'eox_tenant.edxapp_wrapper.backends.oauth_dispatch_j_v1'
+  EOX_TENANT_EDX_AUTH_BACKEND = "eox_tenant.edxapp_wrapper.backends.edx_auth_i_v1"
+
+Those settings can be changed in ``eox_tenant/settings/common.py`` or, for example, in the instance settings.
 
-**Koa (optional)**
+**NOTE**
```
mariajgrimaldi, line 90:
```suggestion
```
In favor of an icon: ℹ️ ?

README.rst
```diff
@@ -1,104 +1,138 @@
...
+| Nutmeg           | >= 6.2         |
++------------------+----------------+
+| Olive            | >= 8.0         |
++------------------+----------------+
+| Palm             | >= v10.0.0     |
++------------------+----------------+
+| Quince           | >= v11.0.0     |
++------------------+----------------+
+
+**NOTE**: Since the 6.2 version, eox-tenant does not support Django 2.2
```
mariajgrimaldi, line 74:
```suggestion
⚠️ The plugin does not support Django 2.2 version ``v6.2.x``.
```

mariajgrimaldi (changes requested):

README.rst
dcoa: You're right

README.rst
dcoa: We use the tenant model but the microsite is still not deprecated

README.rst
dcoa: Please take in mind, that I removed all previous version modifications and listed only the latest (that is why the paragraph mentions the _**the current eox-tenant version**_) to make emphasis (from now and the subsequent releases) the real support for each release, and limit the configuration changes, I mean:

The actual description says 

> If the release you are looking for is not listed, then the accumulation of changes from previous releases is enough.

If I check the changes from Koa to Maple (in the [master branch readme](https://github.com/eduNEXT/eox-tenant/blob/master/README.rst?plain=1#L41)) those are the same, so I'm not accumulating changes, I'm explicitly telling which backends use.

I think using that expression could be confusing, how many previous releases are useful for the accumulation of changes?

README.rst
dcoa: The icon is not supported by .rst so I'm using 🚨 instead.

mariajgrimaldi (dismissed):

README.rst
mariajgrimaldi: Mmm, I get it now that you explained it. Back when I left the comment, I was thinking about the ambiguity of "or the configuration is enough or is not supported." How do I know?

Still, this is a non-blocking comment.

README.rst
```diff
@@ -1,104 +1,136 @@
...
+| Olive            | >= 8.0         |
++------------------+----------------+
+| Palm             | >= v10.0.0     |
++------------------+----------------+
+| Quince           | >= v11.0.0     |
++------------------+----------------+
+
+⚠️ Since the 6.2 version, eox-tenant does not support Django 2.2
+
+The plugin is configured for the latest release (Quince). The following changes in the plugin settings should be applied to be used for previous releases.
```
magajh, line 71:
```suggestion
The plugin is configured for the latest release (Redwood). The following changes in the plugin settings should be applied to be used for previous releases.
```

README.rst
```diff
@@ -1,104 +1,136 @@
...
+
+If the release you are looking for is not listed, the actual configuration is sufficient or it is incompatible with the current eox-tenant version.
```
magajh, line 73: This phrase can be improved for better understanding
```suggestion
If the release you are looking for is not listed, then either the current configuration is sufficient, or the release is incompatible with the current eox-tenant version.
```

README.rst
magajh: @dcoa why are we removing the listed settings for the previous releases and only keeping the latest? 
Could you please explain this decision and the reasoning behind it in the cover letter?

magajh (changes requested):

README.rst
magajh: @dcoa even with these improvements, I'm honestly still having a hard time trying to understand what you explained in the comments and the purpose of this phrase

We can easily check if a release is compatible or not thanks to the [table](https://github.com/eduNEXT/eox-tenant/pull/210/files#diff-7b3ed02bc73dc06b7db906cf97aa91dec2b2eb21f2d92bc5caa761df5bbc168fR48), right? should we specify this here? something like:  _or the release is incompatible with the current eox-tenant version (please refer to the table for compatibillity)._

README.rst
dcoa: @magajh I will explain that here and modify the other repositories once we agree on the best way to write this section.

magajh:
> Proposal
> 
> ➡️ Maintain the settings configuration for the current release.
> 
>  - Simplification: Find the necessary information without being distracted by details that may no longer be relevant (e.g. Ironwood).
>  - Relevance: Users typically work with recent versions of software. Documenting the latest settings ensures that our documentation is pertinent and useful for most users.
>  - Maintenance: We lower the risk of outdated information and reduce the effort required to keep the documentation current. (e.g., before Redwood we have that the current configuration works for Juniper, that is not true because the latest version at that moment doesn't support that release).
>  - Accuracy: During the migration process, we are adding the name of the previous release in the documentation even when the settings are the same, which could lead to unnecessary action for the user.
> 
> Also, the packages management like (pypi or npm - use the readme and the latest release to explain the documentation) if you want to install another version you should select it and that will show you the relevant information for that tag (readme for that version). That behavior is another motivation for this change, showing the information relevant to the version that the user wants to use.

Hi @dcoa,

Thank you for the detailed explanation in the PR description regarding the removal of configurations for previous Open edX releases. Your points about simplification, relevance, maintenance, and accuracy are well noted.

However, I have a few thoughts:

1. If we are including an Open edX named version in the compatibility table, I think we should keep documentation of the settings necessary for successfully using this plugin with that version. That's why I'm not entirely convinced about removing all previous settings.

2. If we are going to direct users to check the release notes, we need to ensure we consistently include the settings for the supported Open edX versions in the release notes of a new eox-core version, or at least when there are breaking changes requiring updates in the settings for a specific release version. This should be documented in our [release strategy](https://github.com/eduNEXT/edunext-internal-documentation/pull/194)

I think it's a good idea to include this information in the release notes from now on. Maybe we could keep the currently listed settings configurations in a document on RTD and just provide a link to it in this README so users can check it if needed?

README.rst
Alec4r: I agree, since the table defines the range of versions supported by a release,

Alec4r:
> > Proposal
> > ➡️ Maintain the settings configuration for the current release.
> > 
> > * Simplification: Find the necessary information without being distracted by details that may no longer be relevant (e.g. Ironwood).
> > * Relevance: Users typically work with recent versions of software. Documenting the latest settings ensures that our documentation is pertinent and useful for most users.
> > * Maintenance: We lower the risk of outdated information and reduce the effort required to keep the documentation current. (e.g., before Redwood we have that the current configuration works for Juniper, that is not true because the latest version at that moment doesn't support that release).
> > * Accuracy: During the migration process, we are adding the name of the previous release in the documentation even when the settings are the same, which could lead to unnecessary action for the user.
> > 
> > Also, the packages management like (pypi or npm - use the readme and the latest release to explain the documentation) if you want to install another version you should select it and that will show you the relevant information for that tag (readme for that version). That behavior is another motivation for this change, showing the information relevant to the version that the user wants to use.
> 
> Hi @dcoa,
> 
> Thank you for the detailed explanation in the PR description regarding the removal of setting configurations for previous Open edX releases. Your points about simplification, relevance, maintenance, and accuracy are well noted.
> 
> However, I have a few thoughts:
> 
>     1. If we are including an Open edX named version in the compatibility table, I think we should keep documentation of the settings necessary for successfully using this plugin with that version. That's why I'm not entirely convinced about removing all previous settings.
> 
>     2. If we are going to direct users to check the release notes, we need to ensure we consistently include the setting config for the supported Open edX versions in the release notes of a new eox-core version, or at least when there are breaking changes requiring updates in the configuration for a specific named version. This should be documented in our [release strategy](https://github.com/eduNEXT/edunext-internal-documentation/pull/194)
> 
> 
> I think it's a good idea to include this information in the release notes from now on. Maybe we could keep the currently listed settings configurations in a document on RTD and just provide a link to it in this README so users can check it if needed?

Hi,

I have been thinking about just adding the link to the tag/documentation should be enough: e.g: https://github.com/eduNEXT/eox-tenant/blob/v6.2.0/README.rst, at least for releases that are not compatible with the current version.

Because if the table show that nutmeg is compatible with versions greater than or equal to 6.2, it means the latest version 11.7.0 is compatible also.

Its assumed that a version is compatible by default if the user doesn't have to change any setting, if the user has to do something, we should add it to our docs.

However, remember that we have "contract" to support the previous release and current release, if a release is 2 version before the current, we could try but is not mandatory.

dcoa:
*To clarify, I will use “version” to talk about eox-tenant and release to refer to Open edX.*

I'm not 100% sure if I caught your concerns but those are my thoughts:

@Alec4r, the contract is still maintained, this repository is a good example of it, the changes for Maple, which is the oldest release supported for 11.x, are listed. Nutmeg, Palm, and Olive are not listed because they use the same configuration as Quince does. 

I was thinking of the major release (for the settings list), and the paragraph description does not explain that, sorry for that.

> Maybe we could keep the currently listed settings configurations in a document on RTD and just provide a link to it in this README so users can check it if needed?

@magajh, I'm not against this idea, even at some point we have to do that because of the length that section would have in the future.

The compatibility notes table should exist as it is because that gives us an idea of which version to install for each Open edX release, we agree about that. However from my point of view the usage section should describe the information for the eox-tenant version that is displayed. If the user wants to use version 6.x, they can visit that tag and the readme gives the corresponding settings.

Let's use Django documentation as an example, when you open it, the latest release (5.0) is displayed if you are using a previous one, you should select the version, that changes the documentation. (I think is the same but inside GitHub).

Of course, the most important here is to have documentation useful for anyone and you have a better understanding of what is necessary, I'm open to making any relevant changes. 😁

README.rst
magajh, line 73: What do you think about adding something like:
```suggestion
If the release you are looking for is not listed, please note:
- If the Open edX release is compatible with the current eox-tenant version (see `Compatibility Notes <https://github.com/eduNEXT/eox-tenant?tab=readme-ov-file#compatibility-notes>`_), the default configuration is sufficient.
- If incompatible, you can refer to the README from the relevant version tag for configuration details (e.g., `v6.2.0 README <https://github.com/eduNEXT/eox-tenant/blob/v6.2.0/README.rst>`_).
```
this way we can explain to the user what to do if the release is not compatible

magajh (approved):
@dcoa @Alec4r thank you both for your comments!

@dcoa I agree that this README is good as it is. I don't think we need to make any more changes, except for one suggestion in the section where we explain what happens if the version the user wants is not listed, for better understanding.

Please ping me if you need my approval again.

magajh (dismissed):

README.rst
```diff
@@ -1,106 +1,139 @@
...
+Those settings can be changed in ``eox_tenant/settings/common.py`` or, for example, in the instance settings.
+
+🚨 If the release you are looking for is not listed, please note:
 
-**Juniper**
+- If the Open edX release is compatible with the current eox-tenant version (see `Compatibility Notes <https://github.com/eduNEXT/eox-tenant?tab=readme-ov-file#compatibility-notes>`_), the default configuration is sufficient.
+- If incompatible, you can refer to the README from the relevant version tag for configuration details (e.g., `v6.2.0 README <https://github.com/eduNEXT/eox-tenant/blob/v6.2.0/README.rst>`_).
 
-For version >= 3.4
+🚨 For version > 10.0.0 you need to enable eox-tenant adding in the LMS configuration:
```
BryanttV, line 90: Just one last comment. I'll be on the lookout to approve the PR
```suggestion
🚨 For version < 10.0.0 you need to enable eox-tenant adding in the LMS configuration:
```

dcoa: 😱 Good catch thank you @BryanttV !!

BryanttV (approved):
LGTM!

magajh (approved):

PR #227, opened 2024-11-06
chore: removed duplicate django constraint

Asespinel:
**Description**

This PR simply erases one duplicate constraint in eox tenant

### DIFF
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -8,7 +8,6 @@
 # pin when possible.  Writing an issue against the offending project and
 # linking to it here is good.
 
-Django<5
 
 # Common constraints for Open edX repos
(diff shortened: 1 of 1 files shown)


BryanttV (approved):

mariajgrimaldi (changes requested):
Running `make upgrade` after this change shouldn't have much impact besides removing this reference: https://github.com/eduNEXT/eox-tenant/blob/602e6b0e8a56805d0e2e5349546a4d9703d228fb/requirements/base.txt#L35. You can ignore other requirement upgrades for the sake of this PR.

mariajgrimaldi (approved):

BryanttV (approved):

PR #246, opened 2025-08-21
feat: added base allowed application

DeimerM:
A base allowed application list has been added to avoid issues with studio and aspects login

### DIFF
--- a/eox_tenant/constants.py
+++ b/eox_tenant/constants.py
@@ -5,7 +5,7 @@ LMS_ENVIRONMENT = getattr(settings, "SERVICE_VARIANT", None) == "lms"
 CMS_ENVIRONMENT = getattr(settings, "SERVICE_VARIANT", None) == "cms"
 LMS_CONFIG_COLUMN = "lms_configs"
 CMS_CONFIG_COLUMN = "studio_configs"
-BASE_ALLOWED_APPLICATIONS = [
+DEFAULT_ALLOWED_AUTH_APPLICATIONS = [
     "cms-sso",
     "cms-sso-dev"
 ]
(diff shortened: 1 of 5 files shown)


eox_tenant/constants.py
```diff
@@ -5,3 +5,9 @@
 CMS_ENVIRONMENT = getattr(settings, "SERVICE_VARIANT", None) == "cms"
 LMS_CONFIG_COLUMN = "lms_configs"
 CMS_CONFIG_COLUMN = "studio_configs"
+BASE_ALLOWED_APPLICATIONS = [
+    "cms-sso",
+    "cms-sso-dev",
+    "superset-sso",
```
magajh, line 11: I'm fine with CMS as the base allowed application, but I don't think aspects should be in the defaults. Many instances don't use aspects, so I think it would be better to keep CMS as the default and make instances explicitly add the setting when they need aspects

magajh (commented):
Thanks @DeimerM

eox_tenant/constants.py
magajh: My reasoning is that CMS/Studio is present in all instances. But Aspects is a separate and optional service, not part of the core platform, and not used by many instances. I think it's cleaner to only configure settings for services that are actually present, rather than having default configurations for optional services that may not be installed

magajh (dismissed):
Let's please fix the name of the constant before merging

Asespinel (dismissed):
LGTM

eox_tenant/constants.py
```diff
@@ -5,3 +5,9 @@
 CMS_ENVIRONMENT = getattr(settings, "SERVICE_VARIANT", None) == "cms"
 LMS_CONFIG_COLUMN = "lms_configs"
 CMS_CONFIG_COLUMN = "studio_configs"
+BASE_ALLOWED_APPLICATIONS = [
```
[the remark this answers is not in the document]
DeimerM: Done!

Asespinel (approved):

Asespinel (approved):

magajh (approved):

PR #262, opened 2026-06-23
feat: make case sensitive name org field

johanseto:
## Description

This PR helps to create multiple organizations based on the course_org_filter sensitive. Also sometimes the org is not found if only once organizations was created. Then if a user creates courses with an org with different  org, the tenant org course not be loaded by the override of get_all_orgs managed by tenant_wise proxy.

eg:

https://github.com/openedx/openedx-platform/blob/release/teak/openedx/core/djangoapps/site_configuration/models.py#L141

because the get_all_org, only return in one case.

https://github.com/eduNEXT/eox-tenant/blob/aca527b44e9a089dfa211548062b23713fa8efd7/eox_tenant/tenant_wise/proxies.py#L109
## Test

Migrate

## After

**[Jira story](https://edunext.atlassian.net/browse/FUTUREX-1778?atlOrigin=eyJpIjoiNDRmMDc1ZmNiODg0NGMyYTg1NTQ3ODRkN2JmNTAyNDAiLCJwIjoiaiJ9)**:

### DIFF
--- a/eox_tenant/migrations/0009_tenant_org_case_name_case_sensitive.py
+++ b/eox_tenant/migrations/0009_tenant_org_case_name_case_sensitive.py
@@ -1,0 +1,20 @@
+# Add collation for tenant organization name column to make it case sensitive.
+
+import collections
+from django.db import migrations, models
+import jsonfield.fields
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('eox_tenant', '0008_synchronize_tenants'),
+    ]
+
+    operations = [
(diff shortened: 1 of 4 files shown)


andrey-canon (approved):
LGTM! Please open the corresponding PR to include this in the main branch.

When you create it, could you please provide a bit more detail in the description? Right now, I'm personally aware of the context and know what the issue is about, but the current description doesn't completely explain why this change is necessary and under which specific cases the failures occur. It would be great to have that documented for future reference.

